PR #161, opened 2019-01-28
Restart

Ugo:
2 changes in this PR:
- There is now a session id for the core, it changes at every restart of the core. This session id is sent back to the worker in every response of the ping.
- An endpoint has been created for the worker to retrieve the list of tasks it is working on.

### DIFF
--- a/src/main/java/com/iexec/core/worker/WorkerController.java
+++ b/src/main/java/com/iexec/core/worker/WorkerController.java
@@ -8,6 +8,7 @@ import com.iexec.common.utils.BytesUtils;
 import com.iexec.common.utils.SignatureUtils;
 import com.iexec.core.chain.ChainConfig;
 import com.iexec.core.chain.CredentialsService;
+import com.iexec.core.configuration.SessionService;
 import com.iexec.core.configuration.WorkerConfiguration;
 import com.iexec.core.security.ChallengeService;
 import com.iexec.core.security.JwtTokenProvider;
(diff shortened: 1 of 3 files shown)


src/main/java/com/iexec/core/worker/WorkerService.java
```diff
@@ -58,6 +59,15 @@ public Worker addWorker(Worker worker) {
         return Optional.empty();
     }
 
+    public List<String> getChainTaskIds(String walletAddress) {
+        Optional<Worker> optional = workerRepository.findByWalletAddress(walletAddress);
+        if (optional.isPresent()) {
+            Worker worker = optional.get();
+            return worker.getComputingChainTaskIds();
```
jeremyjams: Why do we only take 'computingChainTaskIds'?
I think we need them all

src/main/java/com/iexec/core/worker/WorkerController.java
```diff
@@ -124,4 +123,14 @@ public ResponseEntity registerWorker(@RequestHeader("Authorization") String bear
     }
 
 
+    @RequestMapping(method = RequestMethod.GET, path = "/workers/currenttasks")
+    public ResponseEntity<List<String>> getTasksInProgress(@RequestHeader("Authorization") String bearerToken) {
+        String workerWalletAddress = jwtTokenProvider.getWalletAddressFromBearerToken(bearerToken);
+        if (workerWalletAddress.isEmpty()) {
+            return ok(Collections.emptyList());
```
Zied-Guesmi: shouldn't we return unauthorized here?

PR #196, opened 2019-02-12
Send replicate to worker only if it has a recent enough last block nu…

Ugo:
…mber

### DIFF
--- a/src/main/java/com/iexec/core/task/TaskService.java
+++ b/src/main/java/com/iexec/core/task/TaskService.java
@@ -113,7 +113,7 @@ public class TaskService {
 
         for (Task task : runningTasks) {
             String chainTaskId = task.getChainTaskId();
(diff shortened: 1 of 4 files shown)


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -113,7 +113,9 @@ public TaskService(TaskRepository taskRepository,
 
         for (Task task : runningTasks) {
             String chainTaskId = task.getChainTaskId();
-            if (!replicatesService.hasWorkerAlreadyParticipated(chainTaskId, walletAddress) &&
+            boolean blockNumberAvailable = task.getInitializationBlockNumber() == 0 || task.getInitializationBlockNumber() <= blockNumber;
```
jeremyjams: Is it really what we want `task.getInitializationBlockNumber() == 0 ` ?

Ugo: right this should handled before.

PR #199, opened 2019-02-14
Manage pending initializing task

Ugo:


src/main/java/com/iexec/core/chain/IexecHubService.java
```diff
@@ -158,6 +168,28 @@ private long getMaxNbOfPeriodsForConsensus() {
         return Optional.of(Pair.of(chainTaskId, chainReceipt));
     }
 
+    private boolean checkPendingInitialization(String chainDealId, BigInteger taskIndexBigInteger) {
+
+        String computedChainTaskId = ChainUtils.generateChainTaskId(chainDealId, taskIndexBigInteger);
+        long maxWaitingTime = 2 * 60 * 1000L; // 2min
```
jeremyjams: Could `maxWaitingTime` be indexed on the time between two blocks (e.g. 10 blocks in millsec) instead of using hardcoded minutes?

jeremyjams: https://ethereum.stackexchange.com/a/28441

Ugo: right, I fixed it.

PR #203, opened 2019-02-15
Platform metrics

jeremyjams:


Zied-Guesmi:
How about including those metrics under spring [actuator](https://spring.io/guides/gs/actuator-service/) and we can add our custom endpoints

jeremyjams:
Yes, it's just a starter for Eric, he will tell us what he really needs when he will see it

PR #207, opened 2019-02-19
Send result and data repos to workers

Ugo:


Zied-Guesmi:
exactly what we need! thanks

src/main/resources/application.yml
```diff
@@ -40,9 +40,12 @@ tee:
...
-
 resultRepository:
   protocol: ${IEXEC_RESULT_REPOSITORY_PROTOCOL:http}
   ip: ${IEXEC_RESULT_REPOSITORY_IP:localhost}
   port: ${IEXEC_RESULT_REPOSITORY_PORT:18090}
 
+sms:
+  protocol: ${IEXEC_SMS_PROTOCOL:http}
+  ip: ${IEXEC_SMS_IP:localhost}
+  port: ${IEXEC_SMS_PORT:18090}
```
jeremyjams: I think we want 5000 for SMS port

Ugo: right, bad copy/paste, thanks!

PR #209, opened 2019-02-19
isOwnerOfResult or isPublicResult on getResult

jeremyjams:
When getResult on the ResultRepo:
If beneficiary is not set, then result is available to anyone
If beneficiary is set, the downloader should be the beneficiary

There's no more 'requester' involved in the getResult process

### DIFF
--- a/src/main/java/com/iexec/core/result/ResultController.java
+++ b/src/main/java/com/iexec/core/result/ResultController.java
@@ -111,12 +111,12 @@ public class ResultController {
         Authorization auth = authorizationService.getAuthorizationFromToken(token);
 
         boolean isPublicResult = resultService.isPublicResult(chainTaskId, chainId);
-        boolean isOwnerOfResultAndAuthorized = auth != null
+        boolean isAuthorizedOwnerOfResult = auth != null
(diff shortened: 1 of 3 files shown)


src/main/java/com/iexec/core/result/ResultController.java
```diff
@@ -106,21 +106,27 @@ public ResultController(ResultService resultService,
...
+                                            @RequestHeader(name = "Authorization", required = false) String token,
                                             @RequestParam(name = "chainId") Integer chainId) throws IOException {
         Authorization auth = authorizationService.getAuthorizationFromToken(token);
 
-        if (!(authorizationService.isAuthorizationValid(auth) &&
-                resultService.canGetResult(chainId, chainTaskId, auth.getWalletAddress()))) {
-            return new ResponseEntity(HttpStatus.UNAUTHORIZED);
-        }
+        boolean isPublicResult = resultService.isPublicResult(chainTaskId, chainId);
+        boolean isOwnerOfResultAndAuthorized = auth != null
```
Ugo: What about _isAuthorizedOwnerOfResult_ ?

jeremyjams: As you wish!

PR #211, opened 2019-02-20
Add SGX support in the choice of a task for a worker

Ugo:


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -71,6 +73,10 @@ public TaskService(TaskRepository taskRepository,
         return Optional.empty();
     }
 
+    public boolean doesTaskNeedSGX(Task task) {
```
jeremyjams: We will need that method in the worker too, can you please use instead (& update) this mehod https://github.com/iExecBlockchainComputing/iexec-common/blob/002ed4651e87983e375f3f5ceff13657342b8177/src/main/java/com/iexec/common/tee/TeeUtils.java#L14 pleaase ?

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -48,6 +48,8 @@
     private ReplicatesService replicatesService;
     private ApplicationEventPublisher applicationEventPublisher;
 
+    private static final String SGX_TAG = "0x0000000000000000000000000000000000000000000000000000000000000001";
```
jeremyjams: Can you move this to common?

src/test/java/com/iexec/core/task/TaskServiceTests.java
```diff
@@ -38,7 +37,8 @@
     private final static String DAPP_NAME = "dappName";
     private final static String COMMAND_LINE = "commandLine";
     private final long maxExecutionTime = 60000;
-    private final String noSgxTag = "0x0";
+    private final static String NO_SGX_TAG = "0x0000000000000000000000000000000000000000000000000000000000000000";
```
jeremyjams: We already have this value here https://github.com/iExecBlockchainComputing/iexec-common/blob/002ed4651e87983e375f3f5ceff13657342b8177/src/main/java/com/iexec/common/utils/BytesUtils.java#L10

src/main/java/com/iexec/core/task/TaskService.java
Ugo: done

src/main/java/com/iexec/core/task/TaskService.java
Ugo: done

src/test/java/com/iexec/core/task/TaskServiceTests.java
Ugo: changed.

PR #230, opened 2019-03-11
Manage upload to ipfs

Ugo:
The idea here is to publish the results depending on the beneficiary of the task:
- if the beneficiary is null (meaning the result is public), the result will be pushed to the ipfs node defined in the core.
- if the beneficiary is not null, the result will be pushed in the mongo repo of the result repo (same as before)

Some refactoring will be needed to split the replicate update status into several entrypoints instead of a single one at the moment.

### DIFF
--- a/src/main/java/com/iexec/core/replicate/ReplicatesService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicatesService.java
@@ -189,7 +189,7 @@ public class ReplicatesService {
                                       String walletAddress,
                                       ReplicateStatus newStatus,
                                       ReplicateStatusModifier modifier) {
-        updateReplicateStatus(chainTaskId, walletAddress, newStatus, modifier, null);
+        updateReplicateStatus(chainTaskId, walletAddress, newStatus, modifier, null, "");
     }
 
     // in case the task has been modified between reading and writing it, it is retried up to 100 times
@@ -198,7 +198,8 @@ public class ReplicatesService {
                                       String walletAddress,
                                       ReplicateStatus newStatus,
                                       ReplicateStatusModifier modifier,
-                                      ChainReceipt chainReceipt) {
+                                      ChainReceipt chainReceipt,
+                                      String resultLink) {
 
         if (newStatus == ReplicateStatus.RESULT_UPLOADED && !hasResultBeenUploaded(chainTaskId)) {
             log.error("requested updateResplicateStatus to RESULT_UPLOADED when result has not been"
--- a/src/main/java/com/iexec/core/replicate/ReplicatesService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicatesService.java
@@ -312,7 +312,7 @@ public class ReplicatesService {
     }
 
     private Replicate getReplicateWithBlockchainUpdates(Replicate replicate, ChainContributionStatus wishedChainStatus) {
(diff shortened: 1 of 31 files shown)


src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -346,10 +360,24 @@ private boolean isTaskStatusFailOnChain(String chainTaskId, String walletAddress
         }
     }
 
+    boolean isPublicResult(String chainTaskId, Integer chainId) {
```
jeremyjams: Can we move `isPublicResult` to common? (We're gonna need it for the seperate Result Repo)

src/main/java/com/iexec/core/result/ResultService.java
```diff
@@ -103,22 +139,16 @@ boolean isOwnerOfResult(Integer chainId, String chainTaskId, String downloaderAd
         return true;
     }
 
+    boolean isPublicResult(String chainTaskId) {
+        return isPublicResult(chainTaskId, 0);
+    }
+
     boolean isPublicResult(String chainTaskId, Integer chainId) {
```
jeremyjams: We have a duplicate here

src/main/java/com/iexec/core/result/ipfs/IPFSService.java
```diff
@@ -0,0 +1,49 @@
...
+        Multihash filePointer = Multihash.fromBase58(ipfsHash);
+        try {
+            return Optional.of(ipfs.get(filePointer));
+        } catch (IOException e) {
+            log.error("Error when trying to retrieve ipfs object [hash:{}]", ipfsHash);
+        }
+        return Optional.empty();
+    }
+
+    public boolean doesContentExists(String ipfsHash) {
```
jeremyjams: doesContentExist* (no 's')!

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -170,6 +169,16 @@ private int getNbReplicatesWithGivenStatusJustBeforeWorkerLost(String chainTaskI
         return Optional.empty();
     }
 
+    public Optional<Replicate> getReplicateWithResultUploadedStatus(String chainTaskId) {
```
jeremyjams: It seems we have an issue when worker is lost after UPLOADED

jeremyjams: https://github.com/iExecBlockchainComputing/iexec-core/blob/301656cd5c2353d5c131c95c330717bba388715a/src/main/java/com/iexec/core/replicate/ReplicatesService.java#L148

src/main/java/com/iexec/core/replicate/ReplicatesService.java
Ugo: right, moved to common.

src/main/java/com/iexec/core/result/ResultService.java
Ugo: removed.

src/main/java/com/iexec/core/result/ipfs/IPFSService.java
Ugo: thanks, changed.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
Ugo: great catch, thanks! I've modified the code.

PR #234, opened 2019-03-15
Handle disconnection/reconnection of worker and help it recover interrupted replicates

Zied-Guesmi:
- Handle different cases when worker disconnects and tries to recover from it
- Introduce ReplicateSupplyService that handles replicateDemand requests coming from workers
- Introduce RECOVERING replicateStatus: every transition from this task is permitted, transitions to this task come only from "active" (not ..._failed) statuses.


src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -0,0 +1,336 @@
...
+                    ReplicateStatus.RECOVERING, ReplicateStatusModifier.POOL_MANAGER);
+
+            interruptedReplicates.add(interruptedReplicate);
+        }
+
+        return interruptedReplicates;
+    }
+
+    public RecoveryAction getAppropriateRecoveryAction(Task task, Replicate replicate, long blockNumber) {
+        ChainReceipt chainReceipt = new ChainReceipt(blockNumber, "");
```
Ugo: This one is not needed.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -0,0 +1,336 @@
...
+
+            if (!taskService.isConsensusReached(task)) {
+                return RecoveryAction.WAIT;
+            }
+
+            taskExecutorEngine.updateTask(chainTaskId);
+            return RecoveryAction.REVEAL;
+        }
+
+        return null;
```
Ugo: Let's return an optional rather than a null value. This is valid for all recoverReplicate* methods.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -0,0 +1,336 @@
...
+
+        if (task.inContributionPhase()) {
+            return recoverReplicateInContributionPhase(task, replicate, chainReceipt);
+        }
+
+        if (task.getCurrentStatus().equals(TaskStatus.CONTRIBUTION_TIMEOUT)) {
+            return RecoveryAction.ABORT_CONTRIBUTION_TIMEOUT;
+        }
+
+        if (task.getCurrentStatus().equals(TaskStatus.CONSENSUS_REACHED)) {
```
Ugo: Those two ifs can be combined into one.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -0,0 +1,336 @@
...
+     * REVEALED          => RecoveryAction.COMPLETE
+     * !REVEALED         => null
+     */
+
+    private RecoveryAction recoverReplicateIfRevealed(Replicate replicate) {
+        // refresh task
+        Optional<Task> oTask = taskService.getTaskByChainTaskId(replicate.getChainTaskId());
+        if (!oTask.isPresent()) return null;
+
+        return replicate.containsRevealedStatus()
```
Ugo: Nice one but to difficult to read, can we simplify a bit? :)

src/main/java/com/iexec/core/task/TaskExecutorEngine.java
```diff
@@ -24,12 +28,21 @@ public TaskExecutorEngine(TaskService taskService) {
...
 
-    public void updateTask(String chainTaskId) {
+    public CompletableFuture<Boolean> updateTask(String chainTaskId) {
 
         executorMap.putIfAbsent(chainTaskId, ThreadPoolExecutorUtils.singleThreadExecutorWithFixedSizeQueue(1));
 
         Executor executor = executorMap.get(chainTaskId);
-        executor.execute(() -> taskService.tryUpgradeTaskStatus(chainTaskId));
+
+        return CompletableFuture.supplyAsync(() -> taskService.tryUpgradeTaskStatus(chainTaskId), executor)
```
Ugo: Why do we need to return a completableFuture here? It seems that returning nothing is good enough no?

src/main/java/com/iexec/core/replicate/Replicate.java
```diff
@@ -39,6 +40,27 @@ public ReplicateStatus getCurrentStatus() {
         return this.getLatestStatusChange().getStatus();
     }
 
+    @JsonIgnore
+    public ReplicateStatus getLastRelevantStatus() {
```
jeremyjams: I think this method is very dangerous :/

src/main/java/com/iexec/core/replicate/Replicate.java
```diff
@@ -39,6 +40,27 @@ public ReplicateStatus getCurrentStatus() {
...
+                ReplicateStatus.WORKER_LOST,
+                ReplicateStatus.RECOVERING);
+
+        for (int i = statusList.size() - 1; i >= 0; i--) {
+            if (!ignoredStatuses.contains(statusList.get(i))) {
+                return statusList.get(i);
+            }
+        }
+
+        return null;
```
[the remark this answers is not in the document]
Zied-Guesmi: Better yes

src/main/java/com/iexec/core/chain/SignatureService.java
```diff
@@ -38,20 +40,20 @@ String computeAuthorizationHash(String workerWallet, String chainTaskId, String
         return Numeric.toHexString(Hash.sha3(res));
     }
 
-    public ContributionAuthorization createAuthorization(String workerWallet, String chainTaskId, boolean isTrustedExecution) {
+    public Optional<ContributionAuthorization> createAuthorization(String workerWallet, String chainTaskId, boolean isTrustedExecution) {
```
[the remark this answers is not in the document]
Zied-Guesmi: sorry I was lost in the refactoring

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -0,0 +1,336 @@
...
+            Replicate replicate = oReplicate.get();
+
+            boolean isRecoverable = replicate.isRecoverable();
+            if (!isRecoverable) continue;
+
+            RecoveryAction recoveryAction = getAppropriateRecoveryAction(task, replicate, blockNumber);
+            if (recoveryAction == null) continue;
+
+            // generate contribution authorization
+            Optional<ContributionAuthorization> authorization = signatureService.createAuthorization(
```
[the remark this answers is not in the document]
Zied-Guesmi: done

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
Zied-Guesmi: It is actually, because in the case where ReplicateStatus is CONTRIBUTING and the worker contributes onchain without notifying the core should update the status to CONTRIBUTED and tell the worker to WAIT or REVEAL. So when updating the status to CONTRIBUTED we check if it's correct onchain, thus, we need the chainReceipt with the blocknumber.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
Zied-Guesmi: with pleasure

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
Zied-Guesmi: I thought it would be more readable, but it's okey let's change it

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
Zied-Guesmi: of course, also done!

src/main/java/com/iexec/core/task/TaskExecutorEngine.java
Zied-Guesmi: because in some cases I needed to wait for the execution to end.
Exp: ReplicateSupplyService > recoverReplicateInRevealPhase > L259

src/main/java/com/iexec/core/replicate/Replicate.java
Zied-Guesmi: But we need it to find out in what status the replicate was before trying to RECOVER (one or more times) and before being lost.
For instance: this is a typical exp: CONTRIBUTING > WORKER_LOST > RECOVERING > RECOVERING

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
Ugo: Yes, I can see it is used in methods _recoverReplicateInContributionPhase_ and _recoverReplicateInRevealPhase_ but we don't modify anything to this ChainReceipt. We better pass the _blocknumber_ to those methods and create a _new ChainReceipt(blockNumber, "")_ at the lowest level inside those two methods.

Zied-Guesmi: true yeah, I see your point

PR #261, opened 2019-04-16
Fix #252, Integrate SMS in workflow, generate enclaveChallenge for TEE task.

### LINKED ISSUE #252: Update worker infos when changing the machine and keeping the same wallet
Expected behaviour: when we change the worker machine and keep the same wallet, the worker description (cpu, ram, tee) should be updated.

Actual behaviour: Those informations are not changed (we keep the existing ones)

Zied-Guesmi:
- Add spring feign-client to communicate with resultRepo and SMS
- Update worker configuration when reconnecting (fixes #252 )
- Send SMS IP to worker
- Get generated EnclaveChallenge from SMS and send it to worker.
- Use spring-retry to handle failure of http requests.


src/main/java/com/iexec/core/feign/SmsClientWrapper.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.stereotype.Service;
+
+import feign.FeignException;
+import lombok.extern.slf4j.Slf4j;
+
+
+@Slf4j
+@Service
+public class SmsClientWrapper {
```
Ugo: Isn't SmsService better than *Wrapper?

src/main/java/com/iexec/core/feign/ResultRepoClientWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.stereotype.Service;
+
+import feign.FeignException;
+import lombok.extern.slf4j.Slf4j;
+
+
+@Slf4j
+@Service
+public class ResultRepoClientWrapper {
```
Ugo: Should we call it ResultRepoService instead and move it to the correct package?

src/main/java/com/iexec/core/sms/SmsService.java
```diff
@@ -0,0 +1,22 @@
+package com.iexec.core.sms;
+
+import com.iexec.common.utils.BytesUtils;
+import com.iexec.core.feign.SmsClientWrapper;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public class SmsService {
```
Ugo: Probably SmsService and SmsClientWrapper can be merged.

src/main/java/com/iexec/core/feign/SmsClientWrapper.java
Zied-Guesmi: Actually I thought we'd have a wrapper for each feign client so we handle failure, but if that generates a confusion I'll merge it with the service.

src/main/java/com/iexec/core/feign/ResultRepoClientWrapper.java
Zied-Guesmi: Clearer yes, done it !

src/main/java/com/iexec/core/sms/SmsService.java
Zied-Guesmi: 👍

PR #266, opened 2019-04-23
Optionnal workerVersionRequired conf param

jeremyjams:


src/main/java/com/iexec/core/configuration/WorkerConfiguration.java
```diff
@@ -11,4 +11,8 @@
 
     @Value("${workers.askForReplicatePeriod}")
     private long askForReplicatePeriod;
+
+    @Value("${workers.workerVersionRequired}")
+    private String workerVersionRequired;
+
```
Zied-Guesmi: maybe requiredWorkerVersion would be clearer (?)

jeremyjams: Yes, I wasn't sure !

PR #267, opened 2019-04-23
Give the worker a chance to upload result before notifyin another one

Zied-Guesmi:


src/main/java/com/iexec/core/detector/replicate/ReplicateResultUploadTimeoutDetector.java
```diff
@@ -67,15 +71,19 @@ public void detect() {
             log.info("detected replicate with resultUploadTimeout [chainTaskId:{}, replicate:{}, currentStatus:{}]",
             chainTaskId, uploadingReplicate.getWalletAddress(), uploadingReplicate.getCurrentStatus());
 
-            if (uploadingReplicate.getCurrentStatus() == ReplicateStatus.RESULT_UPLOAD_REQUESTED) {
+            if (task.getCurrentStatus() == TaskStatus.RESULT_UPLOAD_REQUESTED) {
```
jeremyjams: I think you'r right with this one, because *_REQUESTED is set by the pool manager, we did have a bug congrats!

src/main/java/com/iexec/core/detector/replicate/ReplicateResultUploadTimeoutDetector.java
```diff
@@ -67,15 +71,19 @@ public void detect() {
...
                 replicatesService.updateReplicateStatus(chainTaskId, uploadingReplicate.getWalletAddress(),
                         ReplicateStatus.RESULT_UPLOAD_REQUEST_FAILED, ReplicateStatusModifier.POOL_MANAGER);
+
+                taskExecutorEngine.updateTask(task.getChainTaskId());
                 return;
             }
 
-            if (uploadingReplicate.getCurrentStatus() == ReplicateStatus.RESULT_UPLOADING) {
+            if (task.getCurrentStatus() == TaskStatus.RESULT_UPLOADING) {
```
jeremyjams: But I'm not sure that one was bug. Why checking the TaskStatus is better than checking the ReplicateStatus (a ReplicateStatus.RESULT_UPLOADING will trigger a TaskStatus.RESULT_UPLOADING, so I would say it doesnt matter if we are checking the status on the task instead of on the replicate)?

Zied-Guesmi:
#268

PR #271, opened 2019-04-24
Request another worker to upload if the first doesn't 

Zied-Guesmi:


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -353,17 +353,26 @@ private void uploadRequested2UploadRequestTimeout(Task task) {
...
+            task.setResultLink(oUploadedReplicate.get().getResultLink());
+            task.setChainCallbackData(oUploadedReplicate.get().getChainCallbackData());
             updateTaskStatusAndSave(task, RESULT_UPLOADED);
             resultUploaded2Finalized2Completed(task);
-        } else if (replicatesService.getNbReplicatesWithCurrentStatus(task.getChainTaskId(), ReplicateStatus.RESULT_UPLOAD_REQUEST_FAILED) > 0 &&
-                replicatesService.getNbReplicatesWithCurrentStatus(task.getChainTaskId(), ReplicateStatus.RESULT_UPLOADING) == 0) {
+            return;
+        }
+
+        boolean noReplicateIsUploading = replicatesService.getNbReplicatesWithCurrentStatus(
```
jeremyjams: We will have a problem if the worker is gone (WORKER_LOST, or worst RECOVERING) after RESUL_UPLOAD (but we can still merge if you bored!), I think we have a method like `getLastButOne` (?)
What do you think?

src/test/java/com/iexec/core/detector/replicate/ReplicateResultUploadTimeoutDetectorTests.java
```diff
@@ -5,6 +5,7 @@
 import com.iexec.core.detector.replicate.ReplicateResultUploadTimeoutDetector;
 import com.iexec.core.replicate.Replicate;
 import com.iexec.core.replicate.ReplicatesService;
+import com.iexec.core.result.repo.proxy.ResultProxyService;
```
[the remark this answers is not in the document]
Zied-Guesmi: :o

src/main/java/com/iexec/core/task/TaskService.java
Zied-Guesmi: No I'm never bored of this :)
If the worker is lost, then that's a good reason to request another upload, but you're right for the recovery case. I changed it in a way that we check if the worker is uploading or recovering to upload, otherwise another worker should upload.
Good point thanks!

PR #285, opened 2019-05-23
Get task access lock

jeremyjams:


src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -84,13 +84,20 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
         }
 
         for (Task task : runningTasks) {
+            String chainTaskId = task.getChainTaskId();
+
+            taskService.initializeTaskAccessForNewReplicate(chainTaskId);
+            if (taskService.isTaskAccessedForNewReplicate(chainTaskId)){
+                continue;//skip task if being accessed
+            }
+            taskService.setTaskAccessedForNewReplicate(chainTaskId, true);//lock task while being accessed
```
Ugo: We better call the method XXXLock explicitly instead of setXXX.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -104,6 +111,7 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
                 if (enclaveChallenge.isEmpty()) continue;
 
                 replicatesService.addNewReplicate(chainTaskId, walletAddress);
+                taskService.setTaskAccessedForNewReplicate(chainTaskId, false);//release task when replicate is created
```
Ugo: We better call the method releaseXXX explicitly instead of setXXX.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Right!

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Thanks!

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -460,4 +462,17 @@ private void resultUploaded2Finalized2Completed(Task task) {
         updateTaskStatusAndSave(task, COMPLETED);
         applicationEventPublisher.publishEvent(new TaskCompletedEvent(task));
     }
+
+    public void initializeTaskAccessForNewReplicate(String chainTaskId) {
+        taskAccessForNewReplicateLock.putIfAbsent(chainTaskId, false);
+    }
+
+    public Boolean isTaskAccessedForNewReplicate(String chainTaskId) {
```
zied-guesmi: I'd got with isTaskBeingAccessedForNewReplicate()

jeremyjams: Done, much cleaner thank you

PR #287, opened 2019-05-24
Block

Ugo:


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -241,6 +245,7 @@ private void running2ConsensusReached(Task task) {
             applicationEventPublisher.publishEvent(ConsensusReachedEvent.builder()
                     .chainTaskId(task.getChainTaskId())
                     .consensus(task.getConsensus())
+                    .blockNumber(web3jService.getLatestBlockNumber())
```
jeremyjams: .blockNumber(task.getConsensusBlock())

Ugo: right, fixed.

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -236,11 +240,13 @@ private void running2ConsensusReached(Task task) {
             // change the the revealDeadline and consensus of the task from the chainTask info
             task.setRevealDeadline(new Date(chainTask.getRevealDeadline()));
             task.setConsensus(chainTask.getConsensusValue());
+            task.setConsensusReachedBlockNumber(web3jService.getLatestBlockNumber());
```
jeremyjams:
```suggestion
            // TODO: Set real consensusBlockNumber
            task.setConsensusReachedBlockNumber(web3jService.getLatestBlockNumber());
```

PR #290, opened 2019-05-31
Notification type for recovery

jeremyjams:


src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -173,43 +173,51 @@ private boolean isFewBlocksAfterInitialization(Task task) {
             String enclaveChallenge = smsService.getEnclaveChallenge(chainTaskId, task.isTeeNeeded());
             if (task.isTeeNeeded() && enclaveChallenge.isEmpty()) continue;
 
-            Optional<RecoveryAction> oRecoveryAction = getAppropriateRecoveryAction(task, replicate, blockNumber);
-            if (!oRecoveryAction.isPresent()) continue;
+            Optional<TaskNotificationType> optionalTaskNotificationType = getAppropriateRecoveryAction(task, replicate, blockNumber);
+            if (!optionalTaskNotificationType.isPresent()) continue;
 
             // generate contribution authorization
             ContributionAuthorization authorization = signatureService.createAuthorization(
```
zied-guesmi: Shouldn't we generate the contribution authorization only in the case of please_contribute notification?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -173,43 +173,51 @@ private boolean isFewBlocksAfterInitialization(Task task) {
...
                     walletAddress, chainTaskId, enclaveChallenge);
 
-            InterruptedReplicateModel interruptedReplicate = InterruptedReplicateModel.builder()
+            //TODO Send only required extras
+            TaskNotificationExtra taskNotificationExtra = TaskNotificationExtra.builder()
                     .contributionAuthorization(authorization)
-                    .recoveryAction(oRecoveryAction.get())
+                    .blockNumber(task.getConsensusReachedBlockNumber())
```
zied-guesmi: We shouldn't send the blockNumber when the notif is pleas_contribute (consensus is not yet reached)

src/main/java/com/iexec/core/task/listener/TaskListeners.java
```diff
@@ -83,7 +84,7 @@ public void onTaskConsensusReached(ConsensusReachedEvent event) {
             notificationService.sendTaskNotification(TaskNotification.builder()
                     .taskNotificationType(TaskNotificationType.PLEASE_REVEAL)
                     .chainTaskId(chainTaskId)
-                    .blockNumber(event.getBlockNumber())
+                    .taskNotificationExtra(TaskNotificationExtra.builder().blockNumber(event.getBlockNumber()).build())
```
zied-guesmi: Why not just TaskNotificationExtra(event.getBlockNumber())

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Didn't want to do it (-> https://github.com/iExecBlockchainComputing/iexec-core/blob/dc24414a9426e1c0284b57ac2ae4197f6581f1d2/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java#L183) but done now :P

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Done

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
zied-guesmi: Ah sorry didn't understand you meant that, thanks!

PR #294, opened 2019-06-07
Exclude tasks that have reached the contribution deadline in the task…

Ugo:
… scheduling

### DIFF
--- a/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 @Slf4j
 @Service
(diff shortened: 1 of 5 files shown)


src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -84,18 +85,23 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
         List<Task> runningTasks = taskService.getInitializedOrRunningTasks();
         if (runningTasks.isEmpty()) {
             return Optional.empty();
         }
 
-        // return empty if the worker already has enough running tasks
-        if (!workerService.canAcceptMoreWorks(walletAddress)) {
-            return Optional.empty();
-        }
+        // filter the Tasks that have reached the contribution deadline
```
jeremyjams:
```suggestion
        // filter the Tasks that have reached the contribution deadline
        //TODO Trigger contributionDeadline detector here
```

Ugo: Thanks, detector trigger has been added.

PR #298, opened 2019-06-10
Detect task that are not in finalized status but that should be

Ugo:


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -446,20 +449,46 @@ private void resultUploaded2Finalized2Completed(Task task) {
...
+        ChainReceipt chainReceipt = optionalChainReceipt.get();
+
+        Optional<ChainTask> oChainTask = iexecHubService.getChainTask(task.getChainTaskId());
+        if (!oChainTask.isPresent()) {
+            return;
+        }
+        chainTask = oChainTask.get();
+
+        if (chainTask.getStatus().equals(ChainTaskStatus.FAILLED)) {
             log.error("Finalize failed [chainTaskId:{} canFinalize:{}, isAfterRevealDeadline:{}, hasAtLeastOneReveal:{}]",
```
jeremyjams: Old log params swap here :)

PR #301, opened 2019-06-13
Add missing details to the replicate update

Ugo:


src/main/java/com/iexec/core/detector/replicate/UnnotifiedAbstractDetector.java
```diff
@@ -77,19 +82,39 @@ void dectectOnchainCompleted(List<TaskStatus> dectectWhenOffchainTaskStatuses,
...
 
+        String wallet = replicate.getWalletAddress();
+
         for (ReplicateStatus statusToUpdate : statusesToUpdate) {
-            replicatesService.updateReplicateStatus(chainTaskId, replicate.getWalletAddress(),
-                    statusToUpdate, ReplicateStatusModifier.POOL_MANAGER);
-        }
-    }
+            // add details to the update if needed
+            switch (statusToUpdate) {
```
jeremyjams: What do you think about moving this block to the real updateReplicateStatus(..) so we can remove the blockNumber from the update method (since we can't trust the worker) ?

PR #302, opened 2019-06-14
Reopening detector

Ugo:


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -318,17 +322,32 @@ public void consensusReached2Reopened(Task task) {
...
+
+    public void reopening2Reopened(Task task) {
+        reopening2Reopened(task, null);
+    }
 
-        task.setConsensus(null);
-        task.setRevealDeadline(new Date(0));
-        updateTaskStatusAndSave(task, REOPENED, optionalChainReceipt.get());
-        updateTaskStatusAndSave(task, INITIALIZED, optionalChainReceipt.get());
+    public void reopening2Reopened(Task task, ChainReceipt chainReceipt) {
```
jeremyjams: I dont think we need `public void reopening2Reopened(Task task, ChainReceipt chainReceipt)` anymore

Ugo: It is now true for few of those methods, I'll refactor that part in a future PR.

PR #307, opened 2019-06-21
Implement full TEE with sgx

zied-guesmi:


src/main/resources/application.yml
```diff
@@ -35,23 +39,24 @@ chain:
...
+
 sms:
   protocol: ${IEXEC_SMS_PROTOCOL:http}
   host: ${IEXEC_SMS_HOST:localhost}
   port: ${IEXEC_SMS_PORT:5000}
 
-ipfs:
-  host: ${IEXEC_IPFS_HOST:127.0.0.1}
-  port: ${IEXEC_IPFS_PORT:5001}
+scone:
```
jeremyjams: Isn't it
cas:
  scone:
?

zied-guesmi: No CAS is a scone service. So CAS under scone (we have also LAS under scone in the worker side)

PR #318, opened 2019-07-22
Whitelist

Ugo:


src/main/java/com/iexec/core/worker/WorkerController.java
```diff
@@ -102,7 +101,7 @@ public ResponseEntity getToken(@RequestParam(name = "walletAddress") String wall
     public ResponseEntity registerWorker(@RequestHeader("Authorization") String bearerToken,
                                          @RequestBody WorkerConfigurationModel model) {
         String workerWalletAddress = jwtTokenProvider.getWalletAddressFromBearerToken(bearerToken);
-        if (workerWalletAddress.isEmpty()) {
+        if (workerWalletAddress.isEmpty() || !workerService.isAllowedToJoin(workerWalletAddress)) {
```
zied-guesmi: Shouldn't we do this in /workers/login instead? If the worker can't log in it is obvious it can't register.

Ugo: Yes right, changed, thanks a lot Zied

PR #319, opened 2019-07-23
Check the last work demand for a worker

Ugo:


src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -41,6 +45,11 @@ public ResponseEntity getAvailableReplicate(@RequestParam(name = "blockNumber")
             return ResponseEntity.status(HttpStatus.UNAUTHORIZED.value()).build();
         }
 
+        if (!workerService.isWorkerAllowedToAskReplicate(workerWalletAddress)){
+            return ResponseEntity.status(HttpStatus.NO_CONTENT.value()).build();
+        }
+        workerService.updateLastReplicateDemand(workerWalletAddress);
```
zied-guesmi: Maybe `updateLastReplicateDemandDate()`

Ugo: yes, that's better. Renamed

PR #320, opened 2019-07-23
Workflow

jeremyjams:


src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -203,41 +204,41 @@ public void updateReplicateStatus(String chainTaskId,
...
-                                      ReplicateStatusModifier modifier, ReplicateDetails details) {
+        public TaskNotificationType updateReplicateStatus(String chainTaskId, String walletAddress, ReplicateStatus newStatus,
+                                                          ReplicateStatusModifier modifier, ReplicateDetails details) {
         ReplicateStatus currentStatus = null;
         Replicate replicate;
         Optional<ReplicatesList> replicates = getReplicatesList(chainTaskId);
         if (!replicates.isPresent()) {
             log.error(getReplicateStatusErrorLog("replicateList missing",chainTaskId, walletAddress, currentStatus, newStatus, modifier));
-            return;
+            return null;
```
Ugo: Let's use Optional instead of returning null.

src/main/java/com/iexec/core/workflow/ReplicateWorkflow.java
```diff
@@ -1,24 +1,35 @@
...
+import java.util.HashMap;
 
+import static com.iexec.common.notification.TaskNotificationType.*;
 import static com.iexec.common.replicate.ReplicateStatus.*;
 
 
 public class ReplicateWorkflow extends Workflow<ReplicateStatus> {
 
     private static ReplicateWorkflow instance;
+    private HashMap<ReplicateStatus, TaskNotificationType> actionMap = new HashMap<>();
```
Ugo: Juste use the Map interface for the declaration of the variable.

jeremyjams: Totally right, thanks

src/main/java/com/iexec/core/replicate/ReplicatesService.java
jeremyjams: Thanks!

PR #323, opened 2019-08-14
Save workflow as json file & add missing statuses

zied-guesmi:


src/main/java/com/iexec/core/workflow/ReplicateWorkflow.java
```diff
@@ -177,4 +175,17 @@ private void setNextActions() {
         setNextAction(COMPLETED, PLEASE_WAIT);
         setNextAction(COMPLETE_FAILED, PLEASE_ABORT);
     }
+
+    /*
+     * Use this to update the json files when transitions
+     * or actions are changed. 
+     */
+    public static void main(String[] args) throws Exception {
```
jeremyjams: Great idea, can we move that to a test ?

zied-guesmi: Done it, but it's just temporary.

PR #330, opened 2019-09-06
Worker should contribute when app/data_downloa_failed & refactor replicate update

zied-guesmi:


src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -216,153 +216,187 @@ public void updateReplicateStatus(String chainTaskId,
...
     }
 
-    private String getReplicateStatusSuccessLog(String chainTaskId, String walletAddress, ReplicateStatus currentStatus, ReplicateStatus newStatus, ReplicateStatusModifier modifier) {
-        return getUpdateReplicateStatusLogMessage("UpdateReplicateStatus succeeded", "", chainTaskId, walletAddress, currentStatus, newStatus, modifier);
+    private boolean verifyStatus(String chainTaskId, String walletAddress, ReplicateStatus status) {
+        switch (status) {
+            case CONTRIBUTED:
+                return iexecHubService.isStatusTrueOnChain(chainTaskId, walletAddress,
+                        ChainContributionStatus.CONTRIBUTED);
+            case CONTRIBUTE_FAILED:
```
jeremyjams: I think we don't want to update to CONTRIBUTE_FAILED & REVEAL_FAILED anymore

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -216,153 +216,187 @@ public void updateReplicateStatus(String chainTaskId,
...
+            log.error("Cannot update replicate, bad wokfow transition {}",
+                    getLogDetails(chainTaskId, replicate, statusUpdate));
             return Optional.empty();
         }
 
-        replicate = updateReplicateFields(chainTaskId, replicate, statusUpdate);
+        boolean canUpdate = true;
+
+        switch (newStatus) {
+            case CONTRIBUTE_FAILED:
```
jeremyjams: We don't want this status anymore

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zied-guesmi: Why not? the worker may not succeed to contribute or reveal (tx issues or not correct hash)!

jeremyjams: As a worker, are you able to provide me a proof of a failed contribution?
If yes, so we can keep this status

zied-guesmi: Well it's the same also for the result upload!

src/main/java/com/iexec/core/replicate/ReplicatesService.java
Ugo: As discussed, we can include CONTRIBUTE_FAILED and REVEAL_FAILED but there should be no update performed on the replicate.

PR #341, opened 2019-11-28
Gpu

Ugo:


src/main/java/com/iexec/core/worker/WorkerService.java
```diff
@@ -177,7 +177,11 @@ public boolean canAcceptMoreWorks(String walletAddress) {
 
     public int getAliveAvailableCpu() {
```
zied-guesmi: What do you think of making this a mongo query directly?

src/main/java/com/iexec/core/worker/WorkerService.java
```diff
@@ -189,8 +193,35 @@ public int getAliveAvailableCpu() {
...
+            if(worker.isGpuEnabled()) {
+                continue;
+            }
             totalCpus+= worker.getCpuNb();
         }
         return totalCpus;
     }
+
+    // We suppose for now that 1 Gpu enabled worker has only one GPU
+    public int getAliveTotalGpu() {
```
zied-guesmi: Also this.

src/main/java/com/iexec/core/worker/WorkerService.java
```diff
@@ -189,8 +193,35 @@ public int getAliveAvailableCpu() {
...
+        int totalGpus = 0;
+        for(Worker worker: getAliveWorkers()) {
+            if (worker.isGpuEnabled()) {
+                totalGpus++;
+            }
+        }
+        return totalGpus;
+    }
+
+    private int aliveAvailableGpu () {
```
zied-guesmi: And this.

PR #371, opened 2020-06-03
Add stdout to replicate

zguesmi:


src/main/java/com/iexec/core/stdout/ReplicateStdout.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class ReplicateStdout {
+
```
jeremyjams, line 13: I would probably add chainTaskId so we keep the context inside the object

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -280,6 +290,14 @@ public void updateReplicateStatus(String chainTaskId,
             return Optional.empty();
         }
 
+        if (statusUpdate.getDetails() != null && statusUpdate.getDetails().getStdout() != null) {
+            if (statusUpdate.getStatus().equals(COMPUTED)) {
+                String stdout = statusUpdate.getDetails().tailStdout().getStdout();
```
jeremyjams, line 295: Do we tail here again in case a tweaked worker doesn't tail?

src/main/java/com/iexec/core/stdout/TaskStdout.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public TaskStdout(String chainTaskId, List<ReplicateStdout> replicateStdoutList) {
+        this.chainTaskId = chainTaskId;
+        this.replicateStdoutList = replicateStdoutList;
+    }
+
+    public boolean containsWalletAddress(String walletAddress) {
+        return replicateStdoutList.stream().anyMatch(
+            e -> e.getWalletAddress().equals(walletAddress)
```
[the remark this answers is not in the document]
zguesmi: It is a `ReplicateStdout` instance. Sorry, I should have used a clearer name. Changing it.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: Exactly!

src/main/java/com/iexec/core/stdout/ReplicateStdout.java
zguesmi: I thought about that but I estimated it redundant.

Issue #384, opened 2020-10-12
Don't try to initialize task with expired deadline


zguesmi:
**Problem description:**
When the core restarts, we try to re-initialize missed tasks without checking if it is post-deadline. When it is restarted with an old blockchain block number it takes a lot of resources to process all of the onchain task history.

**Possible fix:**
Read only tasks with still-not-reached deadline from the db.
https://github.com/iExecBlockchainComputing/iexec-core/blob/8b82c655cfa522a116ecfc001a0fd541a1612562/src/main/java/com/iexec/core/detector/task/InitializedTaskDetector.java#L56

PR #385, opened 2020-10-15
Ignore expired tasks & clean task executor threads after a max duration

zguesmi:


src/main/java/com/iexec/core/task/Task.java
```diff
@@ -47,11 +48,13 @@
     @Version
     private Long version;
 
+    @Indexed(unique = true)
```
jeremyjams, line 51: Thanks!

zguesmi:
#384

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -54,69 +55,106 @@ public DealWatcherService(IexecHubService iexecHubService,
...
-                        chainDeal.getTag());
-                optional.ifPresent(task -> applicationEventPublisher.publishEvent(new TaskCreatedEvent(task.getChainTaskId())));
-            }
-        } catch (Exception e) {
-            e.printStackTrace();
+        ChainDeal chainDeal = oChainDeal.get();
+        // do not process deals after deadline
+        if (!iexecHubService.isBeforeContributionDeadline(chainDeal)) {
+            log.error("Deal has expired [chainDealId:{}, deadline:{}]",
+                    chainDealId, iexecHubService.getChainDealContributionDeadline(chainDeal));
```
jeremyjams, line 111: Maybe in the future we could have at least 2 iexechub services, one for mutable objects and one for immutable objects (here I see we properly save blockchain calls but iexecHubService.getChainDealContributionDeadline(..) lets think the opposite)

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -63,16 +63,45 @@ public TaskService(TaskRepository taskRepository,
...
+            String imageName,
+            String commandLine,
+            int trust,
+            long maxExecutionTime,
+            String tag
+    ) {
+        boolean isPresent = taskRepository
+                .findByChainDealIdAndTaskIndex(chainDealId, taskIndex)
+                .isPresent();
+        if (isPresent) {
```
jeremyjams, line 92: Maybe we could have a full functional-like block here?

jeremyjams (approved):
CustomSingleExecutorThreadFactory :clap: :pray:

src/main/java/com/iexec/core/chain/DealWatcherService.java
zguesmi: Yes, we should get the most out of immutable objects.

src/main/java/com/iexec/core/task/TaskService.java
zguesmi: Done, much cleaner thanks!

PR #386, opened 2020-10-21
Fix websockets

zguesmi:
- Use a dedicated TaskScheduler for websocket heartbeats.
- Use a dedicated TaskScheduler for `@Scheduled` tasks.
- Use a dedicated TaskExecutor for `@Async` tasks.
- `TaskService` is now the entrypoint to update tasks.


src/main/java/com/iexec/core/config/AsyncConfig.java
```diff
@@ -0,0 +1,68 @@
...
+        @Override
+        public void handleUncaughtException(
+            Throwable ex,
+            Method method,
+            Object... params
+        ) {
+            log.error("Exception in async method " +
+                    "[method:{}, params:{}, exception:{}]",
+                    method, params, ex.getMessage());
+            ex.printStackTrace();
```
jeremyjams, line 65: You can pass the exception to the logger instead of printing stack outside
`	logger.error("{}, {}! An exception occurred!", 
  "Hello", 
  "World", 
  new Exception("Custom exception"));`

jeremyjams (approved):

src/main/java/com/iexec/core/config/AsyncConfig.java
zguesmi: :+1:

Issue #419, opened 2021-08-23
Gracefully reply to the worker when a status is already reported


jeremyjams:
Detectors usually found new on-chain status before the worker notify it to the core.
This leads to failures when the worker tries to update its status.
Instead, the core should reply to the worker a `208 Already Reported` response.
Here is a log where such scenario occurs:
```
2021-08-23 07:08:23.845  INFO 1 --- [    Scheduled-7] c.i.c.d.r.UnnotifiedAbstractDetector     : Detected confirmed missing update (replicate) [is:CONTRIBUTING, should:CONTRIBUTED, taskId:0xe78da00789206a613b98647ec4dd165a58fa5468b9aed0c4417ec428a737cf2d]
2021-08-23 07:08:23.868  INFO 1 --- [    Scheduled-7] c.i.core.replicate.ReplicatesService     : Replicate update request [status:CONTRIBUTED, chainTaskId:0xe78da00789206a613b98647ec4dd165a58fa5468b9aed0c4417ec428a737cf2d, walletAddress:0x1a69b2eb604db8eba185df03ea4f5288dcbbd248, details:ReplicateStatusDetails(chainReceipt=ChainReceipt(blockNumber=375, txHash=null), resultLink=null, chainCallbackData=null, errorMessage=null, cause=null, stdout=null)]
2021-08-23 07:08:23.875  INFO 1 --- [    Scheduled-3] com.iexec.core.task.TaskService          : Task already added [chainDealId:0xaa3abf9814ec22ea88fcc30951140ee0eb324976324207f3983810638ed20ae3, taskIndex:0, imageName:iexechub/vanityeth:3.0.0, commandLine:a, trust:1]
2021-08-23 07:08:23.889  INFO 1 --- [    Scheduled-7] c.i.common.chain.Web3jAbstractService    : Verified check [try:1, function:isContributedTrue, args:[0xe78da00789206a613b98647ec4dd165a58fa5468b9aed0c4417ec428a737cf2d, 0x1a69b2eb604db8eba185df03ea4f5288dcbbd248], maxTry:3, msToWait:6000, msPerBlock:1000]
2021-08-23 07:08:23.893  INFO 1 --- [    Scheduled-7] c.i.c.chain.IexecHubAbstractService      : Get worker weight [address:0x1a69b2eb604db8eba185df03ea4f5288dcbbd248, score:0, weight:2]
[...]

PR #420, opened 2021-08-24
 Gracefully reply to the worker when a status is already reported

mcornaton:
Replies to https://github.com/iExecBlockchainComputing/iexec-core/issues/419#issue-977014916


src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -103,6 +103,17 @@ public ReplicatesController(ReplicatesService replicatesService,
         statusUpdate.setDate(new Date());
         statusUpdate.setSuccess(ReplicateStatus.isSuccess(statusUpdate.getStatus()));
 
+        final var replicateStatusUpdateError = replicatesService.canUpdateReplicateStatus(
+                chainTaskId,
+                walletAddress,
+                statusUpdate);
+        if (replicateStatusUpdateError.isPresent()) {
+            if (replicateStatusUpdateError.get() == ReplicateStatusUpdateError.ALREADY_REPORTED) {
+                return status(HttpStatus.ALREADY_REPORTED.value()).build();
```
mcornaton, line 112: This part needs to add a body with a `TaskNotificationType`.

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
```diff
@@ -0,0 +1,34 @@
...
+
+public enum ReplicateStatusUpdateError {
+    UNKNOWN_REPLICATE("Cannot update replicate, could not get replicate [chainTaskId:{}, UpdateRequest:{}]"),
+    BAD_WORKFLOW_TRANSITION("Cannot update replicate, bad workflow transition {}"),
+    ALREADY_REPORTED("Cannot update replicate, status {} already reported."),
+    GENERIC_CANT_UPDATE("Cannot update replicate [UpdateRequest:{}]");
+
+    private final String errorMessageTemplate;
+
+    ReplicateStatusUpdateError(String errorMessageTemplate) {
```
zguesmi, line 27:
```suggestion
    private ReplicateStatusUpdateError(String errorMessageTemplate) {
```

src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -103,6 +104,18 @@ public ReplicatesController(ReplicatesService replicatesService,
         statusUpdate.setDate(new Date());
         statusUpdate.setSuccess(ReplicateStatus.isSuccess(statusUpdate.getStatus()));
 
+        final var replicateStatusUpdateError = replicatesService.canUpdateReplicateStatus(
```
zguesmi, line 107: This is just my opinion, I find it more readable when the explicit type is used.

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.iexec.core.replicate;
+
+public enum ReplicateStatusUpdateError {
+    UNKNOWN_REPLICATE("Cannot update replicate, could not get replicate [chainTaskId:{}, UpdateRequest:{}]"),
```
zguesmi, line 20: The String interpolation here will only work when using `log.info()` which makes them strongly coupled to each other. We can either remove it from here or use something like `getMessage(chainTaskId, updateRequest (...))` with generic string interpolation (`[chainTaskId:%s, UpdateRequest:%s, ....]`).

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -245,40 +241,32 @@ public void updateReplicateStatus(String chainTaskId,
...
         if (oReplicateList.isEmpty() || oReplicateList.get().getReplicateOfWorker(walletAddress).isEmpty()) {
-            log.error("Cannot update replicate, could not get replicate [chainTaskId:{}, UpdateRequest:{}]",
-                    chainTaskId, statusUpdate);
-            return Optional.empty();
+            return Optional.of(ReplicateStatusUpdateError.UNKNOWN_REPLICATE);
         }
 
         ReplicatesList replicatesList = oReplicateList.get();
-        Replicate replicate = replicatesList.getReplicateOfWorker(walletAddress).get();
+        Replicate replicate = replicatesList.getReplicateOfWorker(walletAddress).orElseThrow();
```
zguesmi, line 258: This is already checked above in
```
... || oReplicateList.get().getReplicateOfWorker(walletAddress).isEmpty()
```

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -303,6 +291,53 @@ public void updateReplicateStatus(String chainTaskId,
...
+        final Optional<ReplicateStatusUpdateError> replicateStatusUpdateError =
+                canUpdateReplicateStatus(chainTaskId,walletAddress,statusUpdate);
+        if (replicateStatusUpdateError.isPresent()
+                && Objects.equals(replicateStatusUpdateError.get(),ReplicateStatusUpdateError.UNKNOWN_REPLICATE)) {
+            log.error(ReplicateStatusUpdateError.UNKNOWN_REPLICATE.getErrorMessageTemplate(), chainTaskId, statusUpdate);
+            return Optional.empty();
+        }
+
+        ReplicatesList replicatesList = getReplicatesList(chainTaskId).orElseThrow();
+        Replicate replicate = replicatesList.getReplicateOfWorker(walletAddress).orElseThrow();
```
zguesmi, line 326: These throws are not try/catched in the controller.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -303,6 +291,53 @@ public void updateReplicateStatus(String chainTaskId,
...
+     */
+    @Retryable(value = {OptimisticLockingFailureException.class}, maxAttempts = 100)
+    public Optional<TaskNotificationType> updateReplicateStatus(String chainTaskId,
+                                                                String walletAddress,
+                                                                ReplicateStatusUpdate statusUpdate) {
+        log.info("Replicate update request [status:{}, chainTaskId:{}, walletAddress:{}, details:{}]",
+                statusUpdate.getStatus(), chainTaskId, walletAddress, statusUpdate.getDetailsWithoutStdout());
+
+        final Optional<ReplicateStatusUpdateError> replicateStatusUpdateError =
+                canUpdateReplicateStatus(chainTaskId,walletAddress,statusUpdate);
```
zguesmi, line 318: Why do we make this call two times here and in the controller?

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
mcornaton: enum constructor are private by default. Is there any need to explicitly add this keyword?

zguesmi:
Well done for a first PR, this part is not easy to attack.

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
```diff
@@ -0,0 +1,34 @@
...
+    BAD_WORKFLOW_TRANSITION("Cannot update replicate, bad workflow transition {}"),
+    ALREADY_REPORTED("Cannot update replicate, status {} already reported."),
+    GENERIC_CANT_UPDATE("Cannot update replicate [UpdateRequest:{}]");
+
+    private final String errorMessageTemplate;
```
zguesmi, line 25: I would say simply `message` instead of `errorMessageTemplate` since the context is already an error class.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Sure, but using only `get` triggers a warning in IntelliJ.
By the way, `geŧ` and `orElseThrow` have the same implementation - at least in OpenJDK 11:
```
    public T get() {
        if (value == null) {
            throw new NoSuchElementException("No value present");
        }
        return value;
    }
```
So I prefer to use `orElseThrow` there.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: They won't throw anything as they are already checked in `canUpdateReplicateStatus`. IntelliJ triggers the same warning as before when using only `get`.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Some methods depends on `ReplicateService#updateReplicateStatus` so it should continue to check whether the replicate status could be updated while the controller must know why the status couldn't be updated.
I have found two solutions to this problem:

1. Reformat all the methods depending on `ReplicateService#updateReplicateStatus` so that it can update without checking whether it is a valid update;
2. Make sure everything still works by having twice the test.

First solution would be better but it may lead to new issues if code coverage is not wide. Would you still prefer to go for the first solution?

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
mcornaton: Sure, good catch!

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
zguesmi: Well, your right, please ignore my comment.

src/main/java/com/iexec/core/replicate/ReplicateStatusUpdateError.java
mcornaton: You're right. I'll see what fits best. Thanks!

src/main/java/com/iexec/core/replicate/ReplicatesController.java
mcornaton: I was unsure about this one :laughing: 
I'm not fond of `var` but I think they can make reading easier when correctly used: when type and variable name are the same and long or when there is a `new ExplicitClassName()`. If that's much harder to read for you, I could turn it back to `ReplicateStatusUpdateError`. Just tell me :)

zguesmi: In this specific case, the type of `replicateStatusUpdateError` is not straightforward yes.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: I think it is more a matter of readability. When `orElseThrow()` is used, that implicitly hints about the possibility of the value being empty which is impossible in this case. `get()` however implies that the reader can continue his way peacefully, the value will be there.

zguesmi: I didn't know about IntelliJ warnings though.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: This is an illustration of what I'm trying to say above, `orElseThrow` means possible failure, which in turn means more handling. But I don't know what to do about IntelliJ warnings.

src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -104,10 +105,8 @@ public ReplicatesController(ReplicatesService replicatesService,
         statusUpdate.setDate(new Date());
         statusUpdate.setSuccess(ReplicateStatus.isSuccess(statusUpdate.getStatus()));
 
-        final var replicateStatusUpdateError = replicatesService.canUpdateReplicateStatus(
-                chainTaskId,
-                walletAddress,
-                statusUpdate);
+        final Optional<ReplicateStatusUpdateError> replicateStatusUpdateError =
```
zguesmi, line 108: Thanks :pray:

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: I see. In `ReplicateService#updateReplicateStatus` when we have `ALREADY_REPORTED` can we return `ReplicateWorkflow.getInstance().getNextAction(...)` since "already reported" means the worker can proceed to the following steps of the workflow? In that case, we wouldn't need the verification in the controller. What do you think?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: IntelliJ warnings can be ignored with a comment (`//noinspection OptionalGetWithoutIsPresent`) but I don't think it is good solution when using different IDEs.
By the way, it also triggers a major warning in SonarLint:
image

Would you mind if we keep the `orElseThrow` call with a comment such as `//  "get" could be used there but triggers a warning`?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: I was thinking about this solution, but I'm not sure whether `ReplicateWorkflow.getInstance().getNextAction(...)` would return the correct next action. What if `core` is 2 steps behind `worker`'s status update? It would probably break something.
Do you have any input on this @jeremyjams?

jeremyjams: Core should ask the worker not to do anything else, since another process should already have notified the worker about the next real action.

mcornaton: Which TaskNotificationType denotes "do not do anything else"? Is that "PLEASE_CONTINUE"?

jeremyjams: I guess an appropriate reply would be `PLEASE_WAIT` since "Please wait since you should already have received other orders" looks meaningful.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: Yes, no problem :+1: . Thank you for your patience.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: So I have done a few tests on this point. When using `PLEASE_WAIT`, the worker seems to not do anything more:
```
2021-08-25 13:34:00.484  INFO 1 --- [         task-5] c.i.w.executor.TaskNotificationService   : update replicate response [chainTaskId:0x11f3f6938ff0edf43fc0a11dedb68b9d5802c3a7c261a49d0d74b49c6e48124a, status:CONTRIBUTED, next:PLEASE_WAIT]
2021-08-25 13:34:00.485  WARN 1 --- [         task-5] c.i.w.executor.TaskNotificationService   : [DEBUG] Action: PLEASE_CONTRIBUTE, next action: PLEASE_WAIT
2021-08-25 13:34:00.485  WARN 1 --- [         task-6] c.i.w.executor.TaskNotificationService   : [DEBUG] Action: PLEASE_WAIT, next action: null
2021-08-25 13:34:00.486  WARN 1 --- [         task-6] c.i.w.executor.TaskNotificationService   : No more actions to do [chainTaskId:0x11f3f6938ff0edf43fc0a11dedb68b9d5802c3a7c261a49d0d74b49c6e48124a]
```

Same goes for `PLEASE_CONTINUE`:
```
2021-08-25 13:41:42.224  INFO 1 --- [         task-5] c.i.w.executor.TaskNotificationService   : update replicate response [chainTaskId:0x553f0f4a6666dc1e1242b6f8c9ef7294e305aaf1e36bd942f656fda0e3dc3411, status:CONTRIBUTED, next:PLEASE_CONTINUE]
2021-08-25 13:41:42.224  WARN 1 --- [         task-5] c.i.w.executor.TaskNotificationService   : [DEBUG] Action: PLEASE_CONTRIBUTE, next action: PLEASE_CONTINUE
2021-08-25 13:41:42.225  WARN 1 --- [         task-6] c.i.w.executor.TaskNotificationService   : [DEBUG] Action: PLEASE_CONTINUE, next action: null
2021-08-25 13:41:42.225  WARN 1 --- [         task-6] c.i.w.executor.TaskNotificationService   : No more actions to do [chainTaskId:0x553f0f4a6666dc1e1242b6f8c9ef7294e305aaf1e36bd942f656fda0e3dc3411]
```

I'm not sure of the flow of the worker, however it seems to loop on the `TaskNotificationService#onTaskNotification` method as long as it has actions to do. There's no code to handle `PLEASE_WAIT`, `PLEASE_CONTINUE` or `null` so it seems logic to stop there.
I'll continue to search.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -303,9 +304,62 @@ public void updateReplicateStatus(String chainTaskId,
...
+                && canUpdateReplicateStatus(chainTaskId,walletAddress,statusUpdate).isPresent()) {
             return Optional.empty();
         }
 
+        ReplicatesList replicatesList = getReplicatesList(chainTaskId).orElseThrow();           // "get" could be used there but triggers a warning
+        Replicate replicate = replicatesList.getReplicateOfWorker(walletAddress).orElseThrow(); // "get" could be used there but triggers a warning
+        ReplicateStatus newStatus = statusUpdate.getStatus();
+
+        if (newStatus.equals(CONTRIBUTED)) {
+            int workerWeight = iexecHubService.getWorkerWeight(walletAddress);
```
jeremyjams, line 349: I think I see some code duplication with canUpdateReplicateWeight

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -303,9 +304,62 @@ public void updateReplicateStatus(String chainTaskId,
         }
 
         if (!canUpdate) {
+            log.warn("Cannot update replicate {}",
+                    getStatusUpdateLogs(chainTaskId, replicate, statusUpdate));
+            return Optional.of(ReplicateStatusUpdateError.GENERIC_CANT_UPDATE);
+        }
+
+        return Optional.empty();
```
jeremyjams, line 312: I would change method name (here canUpdate returns a cantStatus) or/and add a "sucess" status for the canUpdate method (instead of empty)

mcornaton: I prefer the second solution. A `NO_ERROR` stratus seems great.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Code duplication has been deleted in the new version. It should be better now! :)
By the way, args are not checked until they are needed. I don't think it is an issue there.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -224,6 +220,97 @@ private boolean isStatusBeforeWorkerLostEqualsTo(Replicate replicate, ReplicateS
...
+    }
+
+    /**
+     * Computes arguments used by {@link ReplicatesService#updateReplicateStatus(String, String, ReplicateStatusUpdate, UpdateReplicateStatusArgs)}.
+     * These arguments can be null or empty, no check is done there.
+     */
+    public UpdateReplicateStatusArgs computeUpdateReplicateStatusArgs(String chainTaskId,
+                                                                      String walletAddress,
+                                                                      ReplicateStatusUpdate statusUpdate) {
+        int workerWeight = iexecHubService.getWorkerWeight(walletAddress);
```
jeremyjams, line 302: Contribution and worker weight are only required for some statuses

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -404,26 +483,25 @@ private boolean verifyStatus(String chainTaskId, String walletAddress, Replicate
         }
     }
 
-    private boolean updateReplicateWeight(String chainTaskId, Replicate replicate) {
+    private boolean canUpdateReplicateWeight(String chainTaskId,
```
jeremyjams, line 486: This method looks a little bit empty, cant we merge it in the calling method?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -258,52 +360,29 @@ public void updateReplicateStatus(String chainTaskId,
...
+                                                                UpdateReplicateStatusArgs updateReplicateStatusArgs) {
         log.info("Replicate update request [status:{}, chainTaskId:{}, walletAddress:{}, details:{}]",
                 statusUpdate.getStatus(), chainTaskId, walletAddress, statusUpdate.getDetailsWithoutStdout());
 
-        Optional<ReplicatesList> oReplicateList = getReplicatesList(chainTaskId);
-        if (oReplicateList.isEmpty() || oReplicateList.get().getReplicateOfWorker(walletAddress).isEmpty()) {
-            log.error("Cannot update replicate, could not get replicate [chainTaskId:{}, UpdateRequest:{}]",
-                    chainTaskId, statusUpdate);
+        if (!Objects.equals(
+                    canUpdateReplicateStatus(chainTaskId, walletAddress, statusUpdate, updateReplicateStatusArgs),
```
jeremyjams, line 369: Do we really need Objects.equals?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: I tend to prefer small methods as their name and Javadoc give precious information without having to read ~10 lines of code. Do you think it could be an issue?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: You're right. As we compare enum member, a simple `==` would work.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: That's fixed. Both of them are used only for `CONTRIBUTED` status.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -366,27 +451,25 @@ private boolean canUpdateToBlockchainSuccess(String chainTaskId, Replicate repli
...
-        ReplicateStatusDetails details = statusUpdate.getDetails();
 
         if (!verifyStatus(chainTaskId, replicate.getWalletAddress(), newStatus)) {
             log.error("Cannot update replicate, status is not correct",
                     getStatusUpdateLogs(chainTaskId, replicate, statusUpdate));
             return false;
         }
 
-        if (details == null || details.getResultLink() == null || details.getResultLink().isEmpty()) {
+        if (updateReplicateStatusArgs.getResultLink() == null || updateReplicateStatusArgs.getResultLink().isEmpty()) {
```
jeremyjams, line 466: StringUtils.isEmpty

src/main/java/com/iexec/core/replicate/ReplicatesService.java
jeremyjams: The method name doesn't express entirely what the method does (eg something like: areResutHashAndWorkerWeightPresent or more generic)

mcornaton: If we go for a so-detailed method name, I think it would be better to separate both checks in new methods - `isResultHashPresent` and `isWorkerWeightPresent` - as they don't have anything in common but their presence for the replicate weight to be updated.
Would that suits you?

jeremyjams: Yes, I like that too.
Can you confirm we need to store the whole chainContribution in updateArgs? (required for resultHash and <?>)

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Thanks! This is fixed in the next commit.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: We could technically store only the `resultHash`. However, the current check is on the whole `chainContribution` object:
```
        if (chainContribution == null) {
            log.error("Failed to get chain contribution [chainTaskId:{}, workerWallet:{}]",
                    chainTaskId, walletAddress);
            return false;
        }
```
We could probably call our new check methods - `isWorkerWeightPresent` and `isChainContributionPresent` - in `computeUpdateReplicateStatusArgs` and throw exceptions in case of bad values but I don't see many Exceptions in our current base code. Do you consider it to be a code smell?
If we don't use exceptions there, I don't think we could tweak the behavior of the code to store only `resultHash` instead of the whole `chainContribution` object.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -404,26 +490,31 @@ private boolean verifyStatus(String chainTaskId, String walletAddress, Replicate
         }
     }
 
-    private boolean updateReplicateWeight(String chainTaskId, Replicate replicate) {
+    private boolean isWorkerWeightPresent(String chainTaskId,
+                                          Replicate replicate,
+                                          UpdateReplicateStatusArgs updateReplicateStatusArgs) {
```
zguesmi, line 495: Since we're at it, isn't it enough to pass `workerWeight` as an argument to this method instead of `UpdateReplicateStatusArgs`?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Are you talking about `chainTaskId`, `replicate`, `statusUpdate` and `updateReplicateStatusArgs`? I don't think they should be checked there. It would be better to check them as soon as they are initialized in `replicatesService`.
However, it was not done before my PR. Do you really think it is necessary?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -404,26 +490,31 @@ private boolean verifyStatus(String chainTaskId, String walletAddress, Replicate
...
         }
+        return true;
+    }
 
-        Optional<ChainContribution> chainContribution =
-                iexecHubService.getChainContribution(chainTaskId, walletAddress);
+    private boolean isChainContributionPresent(String chainTaskId,
+                                               Replicate replicate,
+                                               UpdateReplicateStatusArgs updateReplicateStatusArgs) {
+        ChainContribution chainContribution = updateReplicateStatusArgs.getChainContribution();
```
zguesmi, line 509: Since we're at it, isn't it enough to pass `chainContribution` as an argument to this method instead of `UpdateReplicateStatusArgs`?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -352,12 +438,14 @@ private boolean canUpdateToBlockchainSuccess(String chainTaskId, Replicate repli
...
-            log.error("Cannot update replicate, status is not correct",
+            log.error("Cannot update replicate, status is not correct {}",
                     getStatusUpdateLogs(chainTaskId, replicate, statusUpdate));
             return false;
         }
 
-        if (newStatus.equals(CONTRIBUTED) && !updateReplicateWeight(chainTaskId, replicate)) {
+        if (newStatus.equals(CONTRIBUTED)
+                && (!isWorkerWeightPresent(chainTaskId, replicate, updateReplicateStatusArgs)
+                || !isChainContributionPresent(chainTaskId, replicate, updateReplicateStatusArgs))) {
```
zguesmi, line 448: I like how you split the method. Cleaner!

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: That could be a thing. But the current method name implies that we are looking for the `workerWeight` in an object. To be clearer, we should rephrase it to `isWorkerWeightCorrect` - or a better phrasing if you find one, I'm not 100% satisfied with this one.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Same here, we should rephrase the method name to be clearer :)

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -224,6 +221,103 @@ private boolean isStatusBeforeWorkerLostEqualsTo(Replicate replicate, ReplicateS
...
+                                                                      String walletAddress,
+                                                                      ReplicateStatusUpdate statusUpdate) {
+        UpdateReplicateStatusArgs.UpdateReplicateStatusArgsBuilder builder = UpdateReplicateStatusArgs.builder();
+        if (statusUpdate.getStatus() == CONTRIBUTED) {
+            builder
+                    .workerWeight(iexecHubService.getWorkerWeight(walletAddress))
+                    .chainContribution(iexecHubService.getChainContribution(chainTaskId, walletAddress).orElse(null));
+        }
+
+        ReplicateStatusDetails details = statusUpdate.getDetails();
```
jeremyjams, line 311: Only required for RESULT_UPLOADED

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -404,26 +491,29 @@ private boolean verifyStatus(String chainTaskId, String walletAddress, Replicate
...
             log.error("Failed to get worker weight [chainTaskId:{}, workerWallet:{}]",
                     chainTaskId, walletAddress);
             return false;
         }
+        return true;
+    }
 
-        Optional<ChainContribution> chainContribution =
-                iexecHubService.getChainContribution(chainTaskId, walletAddress);
+    private boolean isChainContributionCorrect(String chainTaskId,
```
jeremyjams, line 506: chainContributionResultHash should not be empty as well

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: I generally tend to use `validate<target>` when it comes to checking variables or states. In this case, it would be `validateWorkerWeight()`. I also don't mind if you keep the method as it is, we can enhance it in coming iterations.

mcornaton: Thanks, `validate<target>` seems better. That's fixed!

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Fixed, thanks!

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Fixed, thanks!

jeremyjams (approved):
:clap: !

PR #427, opened 2021-09-22
TEE task should transition to `FAILED` status when all workers have failed

mcornaton:


src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +432,25 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
         }
     }
 
+    private void running2AllWorkersFailed(Task task) {
+        boolean isRunningTask = task.getCurrentStatus().equals(RUNNING);
+        boolean haveAllWorkerFailed = replicatesService.getReplicates(task.getChainTaskId())
+                .stream()
+                .filter(replicate -> ReplicateStatus.isFailure(replicate.getCurrentStatus()))
+                .count() == workerService.getAliveWorkers().size();
```
jeremyjams, line 440: 1. IMO check is not accurate, because we can have:
worker1replicate *_failed  and not alive anymore
worker2replicate *_failed and alive
worker3replicate *_failed and alive
failed == 3 vs getAliveWorkers == 2

Maybe something like:
for each getAliveWorkers { get task.getReplicateOfWorker; if replicate.isFailure failed ++ }
if failed == aliveWorkers, return Status.ALL_FAILED
which checks all alive workers did participate and failed

2. In addition, it might be interested to return Status.ALL_FAILED_ON_SAME_ERROR and Status.ALL_FAILED_FOR_VARIOUS_ERRORS.
A  Status.ALL_FAILED_ON_SAME_ERROR (if multiple workers) might express the issue isn't caused by workers themselves, but rather by App, Requester..

src/main/java/com/iexec/core/task/TaskStatus.java
```diff
@@ -27,6 +27,7 @@
     INITIALIZED,
     INITIALIZE_FAILED,
     RUNNING,
+    ALL_WORKERS_FAILED,
```
jeremyjams, line 30: RUNNING_FAILED might be more appropriate to keep up with the workflow, it precisely describes an all failed before replicate contributions, whereas ALL_WORKERS_FAILED tends to express it did happen on any other task stage.

src/main/java/com/iexec/core/task/TaskService.java
zguesmi: 1. What about lost workers who contributed ? I think we should first check if at least one worker is in `COMPUTED`, `CONTRIBUTING` or `CONTRIBUTED` status, if it is the case even if the worker is lost we should wait.
2. I'm not sure we need this, it might be redundant since we already have the information under Replicates.

mcornaton: 1. New version has the following logic: a task is transitioned to `RUNNING_FAILED` if:
  1.1. All alive workers have tried to run the task;
  1.2 All workers that have tried to run the task have failed, whether they are alive or lost;
  1.3 No replicate has reached the `COMPUTED` status.

2. I agree with @zguesmi: this piece of information can be retrieved otherwise.

src/main/java/com/iexec/core/task/TaskStatus.java
mcornaton: Fixed!

mcornaton:
New version depends on another PR in `iexec-common`: https://github.com/iExecBlockchainComputing/iexec-common/pull/262
It won't build before this other PR is merged into `develop`.

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +434,63 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
...
+
+        // If all workers that have run the task have failed while running the task, it is definitely a `RUNNING_FAILURE`.
+        boolean allReplicatesFailed = replicates
+                .stream()
+                .map(Replicate::getLastRelevantStatus)
+                // A replicate that has not ever been created nor started is probably failed.
+                .map(status -> status.orElse(ReplicateStatus.FAILED))
+                .filter(Predicate.not(ReplicateStatus::isRunningFailure))
+                .allMatch(ReplicateStatus.FAILED::equals);
+
```
jeremyjams, line 481: The syntax looks tricky here.

Why not getting the replicate of the task for each alive worker (if missing abort)  and checking for each if isFailedBeforeComputed (if not abort at the first)?
(Note sure we should handle offline worker, the question is what we have now)

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +434,63 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
...
+        if (!isRunningTask) {
+            return;
+        }
+
+        final List<Replicate> replicates = replicatesService.getReplicates(task.getChainTaskId());
+
+        final List<Worker> aliveWorkers = workerService.getAliveWorkers();
+
+        // If an alive worker has not run the task, it is not a `RUNNING_FAILURE`.
+        final Predicate<Worker> hasWorkerAlreadyTried = worker -> replicates
```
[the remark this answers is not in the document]
mcornaton: As we are dealing with failures, I find `try` to be more appropriate than `participate`. The second one tends to imply the workers have solved successfully the task while the first one tends to imply the workers may have failed after some tries.
If consistency is crucial there, we could rename the variable. Otherwise, I'd prefer to keep the current one.

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +434,63 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
...
+                .allMatch(hasWorkerAlreadyTried);
+        if (!allAliveWorkersTried) {
+            return;
+        }
+
+        // If all workers that have run the task have failed while running the task, it is definitely a `RUNNING_FAILURE`.
+        boolean allReplicatesFailed = replicates
+                .stream()
+                .map(Replicate::getLastRelevantStatus)
+                // A replicate that has not ever been created nor started is probably failed.
```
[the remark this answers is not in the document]
mcornaton: I'm absolutely not sure about that. I've based my work on the `getLastRelevantStatus` return type. That could be wrong.

However, `Replicate` provides a no-args constructor which does not initialize `statusUpdateList`. This constructor seems to be used only in test cases so not a problem as of now... But that could be an issue in the future. 
We could do the following:
1. Make `getLastRelevantStatus` return a `ReplicateStatus` instead of an `Optional<ReplicateStatus>`;
2. Remove the `@NoArgsConstructor` annotation of `Replicate`;
3. Create a new no-args constructor with a single line to initialize the `statusUpdateList` as it is already done in the other constructor: `this.statusUpdateList.add(poolManagerRequest(CREATED));`.

If it's OK for you, I'll have a look at it in another PR.

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +434,63 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
...
+                .map(status -> status.orElse(ReplicateStatus.FAILED))
+                .filter(Predicate.not(ReplicateStatus::isRunningFailure))
+                .allMatch(ReplicateStatus.FAILED::equals);
```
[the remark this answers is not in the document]
mcornaton: That's not an issue as of now because an `APP_DOWNLOAD_FAILED` and `FAILED` lead to the same result. As you've commented later, the syntax looks tricky so it'll have to be refactored to make it easier to read. :smile:

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -454,41 +451,46 @@ private void running2RunningFailed(Task task) {
         }
 
         final List<Replicate> replicates = replicatesService.getReplicates(task.getChainTaskId());
-
         final List<Worker> aliveWorkers = workerService.getAliveWorkers();
 
-        // If an alive worker has not run the task, it is not a `RUNNING_FAILURE`.
-        final Predicate<Worker> hasWorkerAlreadyTried = worker -> replicates
+        final Function<Worker, Optional<Replicate>> getReplicateForWorker = worker -> replicates
```
jeremyjams, line 456: Much easier to read!
We now have some code duplicate can't we benefit of this existing method https://github.com/iExecBlockchainComputing/iexec-core/blob/98171a1759c18b0c8bac81f0a2490e0d634878c7/src/main/java/com/iexec/core/replicate/ReplicatesList.java#L56 ?

mcornaton: Sure. I didn't know about this function!

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +433,64 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
...
+            return;
+        }
+
+        final ReplicatesList replicatesList = replicatesService.getReplicatesList(task.getChainTaskId()).orElseThrow();
+        final List<Worker> aliveWorkers = workerService.getAliveWorkers();
+
+        final List<Optional<Replicate>> replicatesOfAliveWorkers = aliveWorkers
+                .stream()
+                .map(Worker::getWalletAddress)
+                .map(replicatesList::getReplicateOfWorker)
```
jeremyjams, line 460: The list of optionals doesn't add value, you can filter optional to get a plain list, then compare its size to the alive workers list.

mcornaton: You're right. That's fixed :)

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +433,63 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
         }
     }
 
+    /**
+     * When all workers have failed to run the task, we send the new status as soon as possible.
+     * It avoids a state within which we know the task won't be updated anymore, but we wait for a timeout.
+     * <br>
+     * We consider that all workers are in a `RUNNING_FAILED` status when:
+     * <ol>
+     *     <li>All alive workers have tried to run the task;</li>
```
jeremyjams, line 442: I think you can merge <li/>s

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -432,6 +433,63 @@ private void initializedOrRunning2ContributionTimeout(Task task) {
...
+        if (notAllAliveWorkersTried) {
+            return;
+        }
+
+        // If not all alive workers have failed while running the task, that's not a running failure.
+        boolean notAllReplicatesFailed = replicatesOfAliveWorkers
+                .stream()
+                .map(Replicate::getLastRelevantStatus)
+                .map(Optional::get)
+                .anyMatch(Predicate.not(ReplicateStatus::isFailedBeforeComputed));
```
jeremyjams, line 477: Detail: don't you think the positive way is easier to read? (same for above)

mcornaton: It's easier, but could be a bit slower depending on the workerpool size.
However, for the above test, it wouldn't change anything in time complexity to revert the condition and the variable name. I'll do so.

jeremyjams: I agree, thanks

jeremyjams:
FYI, you can check that https://github.com/iExecBlockchainComputing/iexec-core/blob/a5452aa8604e0835b77f64e0d7168a7d9bc74ba3/src/main/java/com/iexec/core/workflow/ReplicateWorkflow.java#L183

mcornaton:
> FYI, you can check that
> 
> https://github.com/iExecBlockchainComputing/iexec-core/blob/a5452aa8604e0835b77f64e0d7168a7d9bc74ba3/src/main/java/com/iexec/core/workflow/ReplicateWorkflow.java#L183

That looks good. As far as I understand it, replicates should continue to contribute on a task `RUNNING_FAILED`.

jeremyjams:
> > FYI, you can check that
> > https://github.com/iExecBlockchainComputing/iexec-core/blob/a5452aa8604e0835b77f64e0d7168a7d9bc74ba3/src/main/java/com/iexec/core/workflow/ReplicateWorkflow.java#L183
> 
> That looks good. As far as I understand it, replicates should continue to contribute on a task `RUNNING_FAILED`.

Here is what we have in this PR:
```
        updateTaskStatusAndSave(task, RUNNING_FAILED);
        updateTaskStatusAndSave(task, FAILED);
```

mcornaton:
> Here is what we have in this PR:
> 
> ```
>         updateTaskStatusAndSave(task, RUNNING_FAILED);
>         updateTaskStatusAndSave(task, FAILED);
> ```
Well, yes, but I don't get the point. Task is updated to `RUNNING_FAILED` then `FAILED` but it does not mean anything for the workers: they should continue their process. `getNextActionWhenStatusAndCause` and `getNextActionWhenStatus` would still reply with `PLEASE_CONTRIBUTE` to usual replicate failed status update (`APP_DOWNLOAD_FAILED` and `DATA_DOWNLOAD_FAILED`). `COMPUTE_FAILED` status update would imply a `PLEASE_ABORT` status - that's the current behavior, is that what we expect?

src/main/java/com/iexec/core/task/event/TaskComputeFailedEvent.java
```diff
@@ -0,0 +1,28 @@
...
+package com.iexec.core.task.event;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+@Builder
+public class TaskComputeFailedEvent {
```
jeremyjams, line 26: TaskRunningFailedEvent?

mcornaton: Thanks, that's cleaner!

jeremyjams (approved):
Thanks! :ok_hand:

Issue #434, opened 2021-09-30
Task meta status should not include replicate status


mcornaton:
`TaskStatus` should describe a state machine whose states should not be able to go back or to loop on themselves. As of now, this requirement is not met because of some status that describe replicate's status, not task status. E.g. `RESULT_UPLOAD_REQUEST` and `RESULT_UPLOAD_REQUEST_TIMEOUT`.

We should get rid of these status as they add some cumbersome behavior. This operation should be done in 2 steps:
1. Check our test coverage is great and smart.
2. Remove incriminated status and fix tests.

mcornaton:
Unit tests code coverage can be found in following archive. Please have a look at method`updateTaskRunnable` in `com.iexec.core.task.TaskService` (file `ns-14/sources/source-4.html`).
[coverage.zip](https://github.com/iExecBlockchainComputing/iexec-core/files/7259848/coverage.zip)

We basically miss more than half of the transitions. There will be some work to achieve 100% of smart unit tests on this part before we can start step #2.

PR #436, opened 2021-10-01
Isolate task update methods

mcornaton:
First step of https://github.com/iExecBlockchainComputing/iexec-core/issues/434: isolation of task update methods


src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -132,7 +136,7 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
             // no need to ge further if the consensus is already reached on-chain
             // the task should be updated since the consensus is reached but it is still in RUNNING status
             if (taskService.isConsensusReached(task)) {
```
jeremyjams, line 138: Why keeping isConsensusReached in TaskService? Should a now pure TaskService depends on ReplicateService?

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -53,26 +45,16 @@
     private final TaskUpdateRequestManager taskUpdateRequestManager;
     private final IexecHubService iexecHubService;
     private final ReplicatesService replicatesService;
-    private final ApplicationEventPublisher applicationEventPublisher;
-    private final WorkerService workerService;
-    private final BlockchainAdapterService blockchainAdapterService;
 
     public TaskService(
             TaskRepository taskRepository,
             TaskUpdateRequestManager taskUpdateRequestManager,
```
jeremyjams, line 51: Do we have taskUpdateRequestManager usages here?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: Not sure about this one. I was thinking about putting only transitional methods into `TaskUpdateManager`. `isConsensusReached` is still a Task-related methods but does not make a task transition from one state to another so I guessed it was still belonging to `TaskService`.
Do you think it should be in another service?

src/main/java/com/iexec/core/task/TaskService.java
mcornaton: We don't, thanks for the catch!

src/main/java/com/iexec/core/task/TaskUpdateManager.java
```diff
@@ -0,0 +1,612 @@
...
+     */
+    @Override
+    public void onTaskUpdateRequest(String chainTaskId) {
+        log.info("Received task update request [chainTaskId:{}]", chainTaskId);
+        this.updateTaskRunnable(chainTaskId);
+    }
+
+    public CompletableFuture<Boolean> publishUpdateTaskRequest(String chainTaskId) {
+        return taskUpdateRequestManager.publishRequest(chainTaskId);
+    }
```
zguesmi, line 88: Why do we use the same service to publish and consume at the same time? I think `publishUpdateTaskRequest()` is redundant, we should use `TaskUpdateRequestManager` directly to publish an update request.

src/main/java/com/iexec/core/task/TaskUpdateManager.java
```diff
@@ -0,0 +1,612 @@
...
+                             BlockchainAdapterService blockchainAdapterService) {
+        this.taskService = taskService;
+        this.taskRepository = taskRepository;
+        this.taskUpdateRequestManager = taskUpdateRequestManager;
+        this.iexecHubService = iexecHubService;
+        this.replicatesService = replicatesService;
+        this.applicationEventPublisher = applicationEventPublisher;
+        this.workerService = workerService;
+        this.blockchainAdapterService = blockchainAdapterService;
+        this.taskUpdateRequestManager.setRequestConsumer(this);
```
zguesmi, line 74: With the comment below we wouldn't need to do this.

src/main/java/com/iexec/core/task/TaskUpdateManager.java
mcornaton: That was the previous behavior in `TaskService` that I only copied. See: https://github.com/iExecBlockchainComputing/iexec-core/blob/6375324c701cf36b858c7483d716ba34107bc2ee/src/main/java/com/iexec/core/task/TaskService.java#L209

I don't exactly know why it has been designed this way. @jeremyjams, any clue on that?
If there's no reason to that publish/consume behavior to happen there, we could certainly remove it.

zguesmi: Ok, we can keep it that way in this PR, and eventually change it after.

zguesmi:
Sorry, it was not on purpose.

zguesmi (approved):
"Let the purge begin!"

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: 2 points:
A - Why having this pre-transition check in a method (in a dedicated service) and still having other pre-transition checks with booleans without method? (non consistency)
B - This method doesn't need an access to the TaskRepo while other methods do (static-like)

mcornaton: A - It's also used in `ReplicateSupplyService` which is different from other pre-transition checks booleans.
B - You're right.

I see your point and it's probably better for it to be in `TaskUpdateManager`. I'll move it. Thanks!

jeremyjams: OK for ReplicateSupplyService thanks.
We might see if a separate checker would make sense

src/main/java/com/iexec/core/task/TaskUpdateManager.java
jeremyjams: We can either 
(1) Merge TaskUpdateManager & TaskUpdateRequestManager (but testing is trickier)
or
(2) Use TaskUpdateRequestManager as a botleneck/filter (feature) between any client service and the TaskUpdateManager. Doing so, no one but the TaskUpdateRequestManager will be able to talk to the TaskUpdateManager  (to ensures TaskUpdateManager is fed properly)

PR #442, opened 2021-10-11
Fix optimistic locking failure exception

mcornaton:
- Change second detectors phase so they don't conflict with primary detectors
- Fix `@Retryable` annotation on `ReplicatesService#updateReplicateStatus`

### DIFF
--- a/src/main/java/com/iexec/core/detector/replicate/ContributionUnnotifiedDetector.java
+++ b/src/main/java/com/iexec/core/detector/replicate/ContributionUnnotifiedDetector.java
@@ -65,8 +65,8 @@ public class ContributionUnnotifiedDetector extends UnnotifiedAbstractDetector {
     public void detectOnChainChanges() {
         detectOnchainContributedWhenOffchainContributing();
 
-        detectorOccurrence = (detectorOccurrence + 1) % LESS_OFTEN_DETECTOR_FREQUENCY;
-        if (detectorOccurrence == 0) {
+        detectorOccurrence++;
+        if (detectorOccurrence % LESS_OFTEN_DETECTOR_FREQUENCY == 0) {
             detectOnchainContributed();
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -227,6 +227,7 @@ private boolean isStatusBeforeWorkerLostEqualsTo(Replicate replicate, ReplicateS
     /*
      * This implicitly sets the modifier to POOL_MANAGER
      */
+    @Retryable(value = {OptimisticLockingFailureException.class}, maxAttempts = 100)
```
zguesmi, line 230: Why do we need a second retry when we have [this](https://github.com/iExecBlockchainComputing/iexec-core/blob/892bce156bc7698ed00fe868c35e12284b6bb08e/src/main/java/com/iexec/core/replicate/ReplicatesService.java#L260) one?

mcornaton: Done! :)

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -328,6 +334,25 @@ public void updateReplicateStatus(String chainTaskId,
...
+
+    @Recover
+    public void updateReplicateStatus(OptimisticLockingFailureException exception,
+                                      String chainTaskId,
+                                      String walletAddress,
+                                      ReplicateStatus newStatus,
+                                      ReplicateStatusDetails details) {
+        log.error("Could not update replicate status, maximum number of retries reached");
+        exception.printStackTrace();
+    }
```
zguesmi, line 354: I think it's better to have each recovery method directly after its own retryable method.

mcornaton: Sure, thanks!

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -333,6 +356,10 @@ public void updateReplicateStatus(OptimisticLockingFailureException exception,
                                       String chainTaskId,
                                       String walletAddress,
                                       ReplicateStatusUpdate statusUpdate) {
+        logUpdateReplicateStatusRecover(exception);
```
jeremyjams, line 359: Mismatching return type on recovering method, how could it work?

PR #456, opened 2021-11-10
Release version 6.3.0

zguesmi:
* Fix concurrent upload requests.
* Merge abort notifications into one notification with custom cause.


src/main/resources/application.yml
```diff
@@ -31,7 +31,7 @@ cron:
 
 workers:
   askForReplicatePeriod: ${IEXEC_ASK_REPLICATE_PERIOD:5000}
-  requiredWorkerVersion: ${IEXEC_CORE_REQUIRED_WORKER_VERSION:6.1.0} #leave empty will allow any worker version
+  requiredWorkerVersion: ${IEXEC_CORE_REQUIRED_WORKER_VERSION:} #leave empty will allow any worker version
```
jeremyjams, line 34: Do we prefer leaving it empty?

zguesmi: I think it's simpler for testing. We can fix it later when releasing the major version. I think we can also change it to `requiredMinimumWorkerVersion`.

jeremyjams: I agree with the simplification concerning the current usage.
About requiredMinimumWorkerVersion, we should introduce this kind of feature yes :+1:

jeremyjams (approved):

PR #457, opened 2021-11-10
Expose blockchain adapter url

mcornaton:


build.gradle
```diff
@@ -126,7 +126,16 @@ dependencies {
     implementation 'net.jodah:expiringmap:0.5.10'
 
     // awaitility
-    testImplementation 'org.awaitility:awaitility:4.0.1'
+    implementation 'org.awaitility:awaitility:4.0.1'
+
+    // web3j dependency
```
jeremyjams, line 131: ?

mcornaton: The worker was throwing an exception when using a local `iexec-common` build (`ClassNotFoundException`) on Awaitility.

mcornaton:
There's no default value for blockchain adapter URL as I consider that a scheduler whose blockchain adapter URL is unknown should not be running. Is this assumption wrong?

build.gradle
jeremyjams: I understand dev purposes but it should not be present here

jeremyjams:
We can discuss on that, but we need at least the ability to inject such value

build.gradle
mcornaton: Alright, that's reverted. However, `iexec-core` will still depend on transitivity of `iexec-common` dependencies which IMHO is a bad idea.

mcornaton:
I don't get your point on blockchain adapter url value. If you want to inject one, you can use `blockchain-adapter.host` and `blockchain-adapter.port` - both of which having default value by the way. Could you precise your question, please?

build.gradle
mcornaton: Would it suit you if we leave these lines as a comment so that we could debug faster without adding it to the deployed versions?

jeremyjams:
Current host and port are for internal use (ibaa deployed next to core):
https://github.com/iExecBlockchainComputing/iexec-core/blob/9cd44514a242c97e20a9d48917bdd58dee819814/src/main/resources/application.yml#L57
but worker need to hit ibaa from the outside.
As an example, ibaa protocol (for workers) is missing (https).

build.gradle
jeremyjams: I understand for transitive dependencies. We must decide what should be present and what shouldn't. For instance, our target is not having a direct web3j dependency in iExec Core Scheduler.

mcornaton: Are you sure about that last point? We need it there, for example: https://github.com/iExecBlockchainComputing/iexec-core/blob/56452cea9b0db738015411dccd89e6d2ce174282/src/main/java/com/iexec/core/result/ResultService.java#L30
For reference, we currently have 5 classes directly depending on Web3J dependency in iexec-core: `DataSetService`, `IexecHubService`, `SignatureService`, `ResultService` and `WorkerController`.

But I get your point for the other dependencies as they are not directly required by iexec-core.

jeremyjams: Yes, if really required we might only import sub libraries:
```
// https://mvnrepository.com/artifact/org.web3j/crypto
implementation 'org.web3j:crypto:4.7.0'
```

src/main/resources/application.yml
```diff
@@ -54,6 +54,7 @@ chain:
   gasPriceCap: ${IEXEC_GAS_PRICE_CAP:22000000000} #in Wei, will be used for txs if networkGasPrice*gasPriceMultiplier > gasPriceCap
 
 blockchain-adapter:
+  protocol: ${IEXEC_CORE_CHAIN_ADAPTER_PROTOCOL:http}
```
jeremyjams, line 57: Thanks

jeremyjams (approved):
:+1:

zguesmi (approved):

PR #461, opened 2021-11-17
Get chain config from blockchain adapter

mcornaton:


src/main/java/com/iexec/core/chain/adapter/UnauthenticatedBlockchainAdapterClient.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Using an authenticated client makes the anonymous requests fail.
+ * So here is a client for anonymous requests.
+ */
+@FeignClient(
+        name = "UnauthenticatedBlockchainAdapterClient",
+        url = "#{blockchainAdapterClientConfig.url}"
+)
+public interface UnauthenticatedBlockchainAdapterClient {
```
jeremyjams, line 33: What about something like that to avoid multiple classes for same interface:
```
    @Bean
    public BasicAuthRequestInterceptor basicAuthRequestInterceptor() {
        return new BasicAuthRequestInterceptor(clientConfig.getUsername(),
                clientConfig.getPassword()) {

            @Override
            public void apply(RequestTemplate template) {
                if (Collections.singletonList("no-auth")
                        .equals(template.headers().get(AUTHORIZATION))) {
                    template.headers().remove(AUTHORIZATION);
                } else {
                    super.apply(template);
                }
            }
        };
    }
```
@Header("Authorization: 'no-auth' ")
``
?

mcornaton: Well, that was useless. The fault was blockchain adapter's. It should have used `permitAll` method on '/config/chain' instead of 'anonymous'. It's fixed in [my last commit](https://github.com/iExecBlockchainComputing/iexec-blockchain-adapter-api/pull/23/commits/ab4816f820907ad2f1a9cb5280eae611e82c4445).

jeremyjams: Welldone :)

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterClient.java
```diff
@@ -20,11 +20,15 @@
 import com.iexec.common.chain.ChainTask;
 import com.iexec.common.chain.adapter.CommandStatus;
 import com.iexec.common.chain.adapter.args.TaskFinalizeArgs;
+import com.iexec.common.config.PublicChainConfig;
 import feign.FeignException;
 import org.springframework.cloud.openfeign.FeignClient;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
+/**
```
jeremyjams, line 29: You can remove this comment

src/main/java/com/iexec/core/chain/ChainConfig.java
```diff
@@ -28,19 +28,19 @@
 @NoArgsConstructor
 public class ChainConfig {
 
-    @Value("${chain.id}")
+    @Value("#{blockchainAdapterService.publicChainConfig.chainId}")
```
jeremyjams, line 31: Don't you think the app should not boot if publicChainConfig null?

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
```diff
@@ -165,4 +169,27 @@ public BlockchainAdapterService(BlockchainAdapterClient blockchainAdapterClient)
...
+    public PublicChainConfig getPublicChainConfig() {
+        if (publicChainConfig != null) {
+            return publicChainConfig;
+        }
+
+        try {
+            final ResponseEntity<PublicChainConfig> response =
+                    blockchainAdapterClient.getPublicChainConfig();
+            if (response.getStatusCode().is2xxSuccessful() && response.hasBody()) {
+                final PublicChainConfig config = response.getBody();
```
zguesmi, line 184: What's the purpose of this local variable?

src/main/java/com/iexec/core/chain/ChainConfig.java
mcornaton: I think so. What would be the point to boot up if the scheduler can't access the chain?

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterClient.java
mcornaton: Good point, thanks!

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
```diff
@@ -165,4 +169,27 @@ public BlockchainAdapterService(BlockchainAdapterClient blockchainAdapterClient)
...
+     */
+    public PublicChainConfig getPublicChainConfig() {
+        if (publicChainConfig != null) {
+            return publicChainConfig;
+        }
+
+        try {
+            final ResponseEntity<PublicChainConfig> response =
+                    blockchainAdapterClient.getPublicChainConfig();
+            if (response.getStatusCode().is2xxSuccessful() && response.hasBody()) {
```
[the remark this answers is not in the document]
mcornaton: I do think so. A null body should trigger an exception. I'll check that info but it's what I remind of my previous tests.

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
mcornaton: Well... Nothing :)
Thanks!

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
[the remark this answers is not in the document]
mcornaton: So the way Feign deals with `null` bodies is funny. It triggers a `Connection refused`:
```
2021-11-18 08:23:34.789 ERROR 1 --- [           main] c.i.c.c.a.BlockchainAdapterService       : Failed to get public chain config:
feign.RetryableException: Connection refused (Connection refused) executing GET http://blockchain-adapter:13010/config/chain
```

On another point, when digging into Feign implementation, we can see that the only branch that leads to a `null` return is when we only expect a `void` return type. So I think we don't have any issue there :)

src/main/java/com/iexec/core/chain/ChainConfig.java
```diff
@@ -28,19 +28,19 @@
...
-    @Value("${chain.sidechain}")
+    @Value("#{blockchainAdapterService.publicChainConfig.isSidechain()}")
     private boolean isSidechain;
 
     @Value("${chain.privateAddress}")
     private String privateChainAddress;
 
-    @Value("${chain.publicAddress}")
+    @Value("#{blockchainAdapterService.publicChainConfig.nodeAddress}")
     private String publicChainAddress;
```
jeremyjams, line 41: I think we don't need that value anymore

mcornaton: Good point, thanks!

zguesmi (approved):
Sounds good to me!

jeremyjams (approved):
:+1:

PR #462, opened 2021-11-18
Feature/exclude junit4

jbern0rd:
Tests updated
`grep -r org.junit src | grep -v jupiter` returns an empty set
(no code uses an org.junit package which does not belong to jupiter, ie junit 5)


jeremyjams (approved):
Thanks for upgrade and wording

PR #463, opened 2021-11-18
Feature/add maven local repository

jbern0rd:
add mavenLocal() before PR on iexec-common with maven-publish

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,46 +70,46 @@ dependencyManagement {
 
 dependencies {
     // iexec
-    compile "com.iexec.common:iexec-common:$iexecCommonVersion"
-    //compile files("../iexec-common/build/libs/iexec-common-${iexecCommonVersion}.jar")
+    implementation "com.iexec.common:iexec-common:$iexecCommonVersion"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -28,6 +28,7 @@ targetCompatibility = 11
 
 
 repositories {
+    mavenLocal()
```
jeremyjams, line 31: :+1:

build.gradle
```diff
@@ -69,46 +70,46 @@ dependencyManagement {
 
 dependencies {
     // iexec
-    compile "com.iexec.common:iexec-common:$iexecCommonVersion"
-    //compile files("../iexec-common/build/libs/iexec-common-${iexecCommonVersion}.jar")
+    implementation "com.iexec.common:iexec-common:$iexecCommonVersion"
+    //implementation files("../iexec-common/build/libs/iexec-common-${iexecCommonVersion}-all.jar")
```
jeremyjams, line 74: Cant we remove that now?

jbern0rd: I prefer to keep it for the moment until the PR in iexec-common is merged and everyone has synchronized its branches.
I think it's best to delete it at a later date. Some other cleanups should be possible at the same time.

zguesmi (approved):

PR #465, opened 2021-11-24
Fix race condition on task attribution

mcornaton:
When a worker looks for a new task to compute, it can take a task that's already computed by another worker currently transitioning its replicate from `CONTRIBUTING` to `CONTRIBUTED`.
That means this worker's weight would never be considered as we retrieve it in two different states in `ContributionService` - once considered as `CONTRIBUTING` and once as `CONTRIBUTED`, but never with the correct status regarding the current filter. The scheduler then consider a new replicate is needed to compute this task... But when the new replicate is started, the first replicate has finished its transition so it is now in `CONTRIBUTED` status. The contribution and pending weights are computed again... And the task comes out of the `RUNNING` status, so all other replicates are canceled - what is called `TASK_NOT_ACTIVE` in worker's logs.


src/main/java/com/iexec/core/contribution/ContributionService.java
```diff
@@ -18,33 +18,27 @@
...
 import com.iexec.core.replicate.Replicate;
-import com.iexec.core.replicate.ReplicatesService;
 import org.springframework.stereotype.Service;
 
 import java.util.HashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
 @Service
```
jeremyjams, line 28: I don't think we need a service anymore, a ContributionHelper is enought.

zguesmi: Same for `PredictionService`.

zguesmi (approved):
Thanks!

jeremyjams (approved):

PR #468, opened 2021-11-25
Add findByCurrentStatus method with sort parameter to order by contributionDeadline

jbern0rd:
Draft PR for comments
To be complete, this feature would merit a basic integration test to check the mongodb query is correct and returns tasks in the expected order.

### DIFF
--- a/src/main/java/com/iexec/core/chain/DealWatcherService.java
+++ b/src/main/java/com/iexec/core/chain/DealWatcherService.java
@@ -25,7 +25,9 @@ import com.iexec.core.task.event.TaskCreatedEvent;
 import io.reactivex.disposables.Disposable;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
 import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.event.EventListener;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
@@ -59,11 +61,12 @@ public class DealWatcherService {
     }
 
     /**
-     * This should be non-blocking to librate
+     * This should be non-blocking to liberate
      * the main thread, since deals can have
      * a large number of tasks (BoT).
      */
     @Async
+    @EventListener(ApplicationReadyEvent.class)
     public void run() {
         subscribeToDealEventFromOneBlockToLatest(configurationService.getLastSeenBlockWithDeal());
     }
--- a/src/main/java/com/iexec/core/task/TaskRepository.java
+++ b/src/main/java/com/iexec/core/task/TaskRepository.java
@@ -16,6 +16,7 @@
 
 package com.iexec.core.task;
 
+import org.springframework.data.domain.Sort;
 import org.springframework.data.mongodb.repository.MongoRepository;
 import org.springframework.data.mongodb.repository.Query;
 
@@ -39,6 +40,9 @@ interface TaskRepository extends MongoRepository<Task, String> {
 
     @Query("{ 'currentStatus': {$in: ?0} }")
     List<Task> findByCurrentStatus(List<TaskStatus> statuses);
+
+    @Query("{ 'currentStatus': {$in: ?0} }")
+    List<Task> findByCurrentStatus(List<TaskStatus> statuses, Sort sort);
 
     @Query("{ 'currentStatus': {$nin: ?0} }")
     List<Task> findByCurrentStatusNotIn(List<TaskStatus> statuses);
--- a/src/main/java/com/iexec/core/task/TaskService.java
+++ b/src/main/java/com/iexec/core/task/TaskService.java
@@ -111,7 +111,7 @@ public class TaskService {
 
     public List<Task> getInitializedOrRunningTasks() {
         return taskRepository.findByCurrentStatus(Arrays.asList(INITIALIZED, RUNNING),
-                Sort.by(Sort.Direction.ASC, "contributionDeadline"));
+                Sort.by(Sort.Direction.ASC, Task.CONTRIBUTION_DEADLINE_FIELD_NAME));
     }
 
     public List<Task> getTasksInNonFinalStatuses() {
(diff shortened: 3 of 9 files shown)


src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -40,6 +41,9 @@
     @Query("{ 'currentStatus': {$in: ?0} }")
     List<Task> findByCurrentStatus(List<TaskStatus> statuses);
 
+    @Query("{ 'currentStatus'; {$in: ?0} }")
```
zguesmi, line 44:
```suggestion
    @Query("{ 'currentStatus': {$in: ?0} }")
```

src/test/resources/logback.xml
```diff
@@ -0,0 +1,34 @@
...
+
+<configuration>
+
+    <!-- Default -->
+    <property name="LOG_LEVEL" value="${IEXEC_LOG_LEVEL:-INFO}"/>
+    <include resource="org/springframework/boot/logging/logback/base.xml"/>
+    <logger name="org.springframework.web" level="${LOG_LEVEL}"/>
+
+    <!-- iexec -->
+    <logger name="com.iexec" leve="DEBUG"/>
```
mcornaton, line 26: Shouldn't this be `level` instead of `leve`?

jbern0rd: Fixed, thanks

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -59,11 +61,12 @@ public DealWatcherService(IexecHubService iexecHubService,
     }
 
     /**
-     * This should be non-blocking to librate
+     * This should be non-blocking to liberate
      * the main thread, since deals can have
      * a large number of tasks (BoT).
      */
     @Async
+    @EventListener(ApplicationReadyEvent.class)
```
jeremyjams, line 69: Nice!

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -109,7 +110,8 @@ public TaskService(TaskRepository taskRepository) {
     }
 
     public List<Task> getInitializedOrRunningTasks() {
-        return taskRepository.findByCurrentStatus(Arrays.asList(INITIALIZED, RUNNING));
+        return taskRepository.findByCurrentStatus(Arrays.asList(INITIALIZED, RUNNING),
+                Sort.by(Sort.Direction.ASC, "contributionDeadline"));
```
jeremyjams, line 114: How about something like:
`Sort.by(Sort.Direction.ASC, Task.CONTRIBUTION_DEADLINE_FIELD_NAME)`
```
@Field(CONTRIBUTION_DEADLINE_FIELD_NAME)
private Date contributionDeadline;
```
 
 To protect against future updates?

zguesmi (approved):
Well done :clap:

src/main/java/com/iexec/core/task/TaskService.java
jbern0rd: Fixed, thanks for the suggestion !

jeremyjams (approved):
Great, thank you for your tests!

mcornaton (approved):
Thanks!

src/test/java/com/iexec/core/task/TaskRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+            Task task = getStubTask(maxExecutionTime);
+            task.setChainDealId(generateChainId());
+            task.setChainTaskId(generateChainId());
+            task.setCurrentStatus(generator.nextInt(50) % 2 == 0 ? RUNNING : INITIALIZED);
+            int amountToAdd = generator.nextInt(10);
+            task.setContributionDeadline(Date.from(Instant.now().plus(amountToAdd, ChronoUnit.MINUTES)));
+            tasks.add(task);
+        }
+        taskRepository.saveAll(tasks);
+        tasks.sort(Comparator.comparing(Task::getCurrentStatus, Comparator.reverseOrder())
```
jeremyjams, line 101: I think it is interesting/clearer to add at least 1 test where expected task list is explicitly defined.

src/test/java/com/iexec/core/task/TaskRepositoryTest.java
```diff
@@ -0,0 +1,111 @@
...
+                .hasCauseExactlyInstanceOf(com.mongodb.MongoBulkWriteException.class)
+                .hasMessageContaining("duplicate key error collection: iexec.task index: chainTaskId dup key");
+    }
+
+    @Test
+    void shouldFindTasksOrderedByCurrentStatusAndContributionDeadline() {
+        List<Task> tasks = new ArrayList<>();
+        for (int i = 0; i < 10; i++) {
+            Task task = getStubTask(maxExecutionTime);
+            task.setChainDealId(generateChainId());
```
jeremyjams, line 93:
```suggestion
            task.setChainDealId(generateHexId());
```

src/test/java/com/iexec/core/task/TaskRepositoryTest.java
jbern0rd: Would it be alright to have a simple test with fixed data and a fuzzy one with randomization to try different combinations ?

jeremyjams: That would be perfect, having both benefits!

src/test/java/com/iexec/core/task/TaskRepositoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+
+import static com.iexec.core.task.TaskStatus.INITIALIZED;
+import static com.iexec.core.task.TaskStatus.RUNNING;
+import static com.iexec.core.task.TaskTestsUtils.getStubTask;
+
+@DataMongoTest
```
mcornaton, line 25: Are we sure we want Spring tests in our unit tests?

jbern0rd: Those are specific tests which only create beans for entities and repositories. The beans annotated with @Components and @Services are not created.
The new tests specifically target and validate the data model which should be validated as early as possible.
There is a valid discussion on the topic and we decided to start this way and maybe to refactor later.

PR #472, opened 2021-12-03
Feature/new oci images conventions

jbern0rd:


jbern0rd:
See https://github.com/iExecBlockchainComputing/iexec-core/pull/473

PR #474, opened 2021-12-08
Fix `initialized2Running`

mcornaton:
- it should transition for all replicate running status to avoid soft locks


mcornaton:
Should be on `master`, not `develop`

PR #479, opened 2021-12-16
Prioritize task update based on their status

mcornaton:


mcornaton:
If we ever need it, this could be easily updated to add deadlines ordering with `thenComparing` in `createQueue` method:
https://github.com/iExecBlockchainComputing/iexec-core/blob/d4a16b6b9caa34ef2415b543c09ab629f1daef21/src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java#L162

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -89,35 +109,60 @@ public void setRequestConsumer(final TaskUpdateRequestConsumer consumer) {
...
             log.error("The unexpected happened", e);
             Thread.currentThread().interrupt();
+        } finally {
+            updatingThreads.decrementAndGet();
         }
     }
+
+    PriorityBlockingQueue<Task> createQueue() {
+        // Tasks whose status are the more advanced should be computed before others
+        final Comparator<Task> comparator = Comparator.comparing(task -> task.getCurrentStatus().ordinal(), Comparator.reverseOrder());
```
jeremyjams, line 162: Why sorting them by status instead of contribution deadline?

mcornaton: When we work with bag of tasks, every task has the same `contributionDeadline`... But some of them may be more advanced in their statuses, so they have priority to be updated.
I agree with you on the fact that sorting task by contribution deadline may be great and that's something that should not be too hard to implement. See https://github.com/iExecBlockchainComputing/iexec-core/pull/479#issuecomment-995699012.

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
```diff
@@ -105,4 +146,36 @@ public void shouldNotUpdateAtTheSameTime() throws NoSuchFieldException, IllegalA
             foundOutputsForKeyGroup.merge(updateId, 1, (currentValue, defaultValue) -> currentValue + 1);
         }
     }
+    // endregion
+
+    // region createQueue()
+    @Test
+    void shouldGetInOrder() {
```
jeremyjams, line 153: I think contribution deadline is missing in this test.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -89,35 +109,61 @@ public void setRequestConsumer(final TaskUpdateRequestConsumer consumer) {
...
             Thread.currentThread().interrupt();
+        } finally {
+            updatingThreads.decrementAndGet();
         }
     }
+
+    PriorityBlockingQueue<Task> createQueue() {
+        // Tasks whose status are the more advanced should be computed before others
+        final Comparator<Task> comparator = Comparator.comparing((Task task) -> task.getCurrentStatus().ordinal(), Comparator.reverseOrder())
+                .thenComparing(Task::getContributionDeadline, Comparator.reverseOrder());
```
jbern0rd, line 163: Shouldn't it be `Comparator.naturalOrder()` for **contributionDeadline** ?
I mean we want to prioritize tasks with a short contribution deadline and need to check we have the correct behaviour.
Second question, do we prioritize by status, then deadline or the other way around ?

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
jbern0rd: I agree with this

jbern0rd: I agree with this

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: You're absolutely right! The condition has been reversed. Thanks!

I think we should prioritize by status then deadline because there could be an attack otherwise: a big task could be never transitioned if lots of small tasks are asked. Then, the big task would fall into contribution timeout, although it has been computed by one or more workers.

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
mcornaton: Sure, thanks for pointing it out. Fixed in https://github.com/iExecBlockchainComputing/iexec-core/pull/479/commits/af0e77ceb3352628c76f91f1798c2f984abbbb2e.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
jbern0rd: Thanks, I get the idea !
After reading https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Enum.html#compareTo(E),
I think the first comparison could be written `Comparator.comparing(Task::getCurrentStatus, Comparator.reverseOrder())`
There is no requirement to explicitly call `ordinal()` as it is handled through the `compareTo()` method.
See also: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Comparable.html

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
```diff
@@ -105,4 +146,107 @@ public void shouldNotUpdateAtTheSameTime() throws NoSuchFieldException, IllegalA
...
+        Collections.shuffle(tasks);
+        System.out.println(tasks);
+        queue.addAll(tasks);
+
+        final List<Task> prioritizedTasks = new ArrayList<>();
+        queue.drainTo(prioritizedTasks);
+        System.out.println(prioritizedTasks);
+
+        Assertions.assertThat(prioritizedTasks)
+                .containsExactlyInAnyOrder(
```
jeremyjams, line 213: In any order?

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: Absolutely! Fixed in https://github.com/iExecBlockchainComputing/iexec-core/pull/479/commits/1cd00c4723e626da108e1586e1814a2f0fca79ee :)

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
mcornaton: Auto-completion... :laughing: 
Fixed in https://github.com/iExecBlockchainComputing/iexec-core/pull/479/commits/1cd00c4723e626da108e1586e1814a2f0fca79ee, thanks!

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -89,35 +109,62 @@ public void setRequestConsumer(final TaskUpdateRequestConsumer consumer) {
...
             log.warn("Waiting for consumer before consuming [queueSize:{}]", queue.size());
             return;
         }
 
-        while (!Thread.currentThread().isInterrupted()){
-            waitForTaskUpdateRequest();
+        // We should only start threads to complete the thread pool.
+        // This is useless for a fresh start but could be handy
+        // if current thread is stopped and current method is rescheduled.
+        final int nbThreadsToStart = TASK_UPDATE_THREADS_POOL_SIZE - updatingThreads.get();
```
jeremyjams, line 122: Step which takes a lot of time is `synchronized { consumer.onTaskUpdateRequest(..) }`, this why we were delegating it to a dedicated executor handling multiple updates at a time. Why now having multiple threads waiting-a-lot then racing-a-lot for next element of the queue?

jeremyjams: Thanks. I think your comment justifies the fact of putting  `queue.take()`  inside the async code block delegated to the pool executor, but why would we need 4 listening threads?

mcornaton: That could maybe be something like: when starting a new update, check if we have some spare thread and start a new `queue.take()` if we do. At the end of each update, check if we still have at least one waiting thread. If we don't, start a new thread.
Does it look good to you?

jeremyjams: Don't we already have 1 waiting thread?

mcornaton: I think there's a misunderstanding. I'm talking about the waiting threads blocked on `queue.take()`. They represent `TASK_UPDATE_THREADS_POOL_SIZE - updatingThreads` threads, waiting for a new task update request to be published.

jeremyjams:
Why not something like:
```
    @Scheduled(fixedDelay = 1000)
    public void consumeAndNotify() {
        if (consumer == null){
            log.warn("Waiting for consumer before consuming [queueSize:{}]", queue.size());
            return;
        }

        while (!Thread.currentThread().isInterrupted()){
            waitForTaskUpdateRequest();
        }
    }

    private void waitForTaskUpdateRequest() {
        log.info("Waiting requests from publisher [queueSize:{}]", queue.size());
        try {
            if (taskUpdateExecutor.getActiveCount() < TASK_UPDATE_THREADS_POOL_SIZE){
                Task task = queue.take();
                CompletableFuture.runAsync(() -> {
                    synchronized (locks.computeIfAbsent(task.getChainTaskId(), key -> new Object())){ // require one update on a same task at a time
                        consumer.onTaskUpdateRequest(task.getChainTaskId()); // synchronously update task
                    }
                }, taskUpdateExecutor);
            }
        } catch (InterruptedException e) {
            log.error("The unexpected happened", e);
            Thread.currentThread().interrupt();
        }
    }
```
?

mcornaton:
@jeremyjams: If you already have `TASK_UPDATE_THREADS_POOL_SIZE` updating threads, you'll trigger a lot of calls to `waitForTaskUpdateRequest` and load the Scheduler... Which is what I'm trying to avoid there.

jbern0rd:
Implementation to discuss:
* publishRequest changes to accept a chainTaskId as a parameter. This id is written in a Queue/List, ...
* We use a method to poll on the list (wait on empty list, work on requests)
* We add in taskRepository a method `List<Task> findByChainTaskId(List<String ids, Sort sort)`
* The new method is used to directly fetch the tasks with the correct order
* The requests update are handled by following the order of tasks returned by the new method
* chainTaskIds from handled tasks are removed from the list

This may allow us to remove some bottlenecks and scheduling problems

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -50,17 +45,22 @@
...
+
+    final TaskUpdatePriorityBlockingQueue queue = new TaskUpdatePriorityBlockingQueue();
+    // Both `corePoolSize` and `maximumPoolSize` should be set to `TASK_UPDATE_THREADS_POOL_SIZE`.
+    // Otherwise, `taskUpdateExecutor` won't pop `maximumPoolSize` threads
+    // as new threads are popped only if the queue is full
+    // - which never happens with an unbounded queue.
+    final ThreadPoolExecutor taskUpdateExecutor = new ThreadPoolExecutor(
+            TASK_UPDATE_THREADS_POOL_SIZE,
+            TASK_UPDATE_THREADS_POOL_SIZE,
+            60,
```
jeremyjams, line 60: * @param keepAliveTime when the number of threads is greater than
     *        the core, this is the maximum time that excess idle threads
     *        will wait for new tasks before terminating.
    
 Why do we need that?

src/main/java/com/iexec/core/task/update/TaskUpdate.java
```diff
@@ -0,0 +1,96 @@
...
+import com.iexec.core.task.TaskStatus;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Comparator;
+import java.util.Date;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentMap;
+
+@Slf4j
+class TaskUpdate implements Runnable, Comparable<TaskUpdate> {
```
jeremyjams, line 29: Nice :)

src/main/java/com/iexec/core/task/update/TaskUpdatePriorityBlockingQueue.java
```diff
@@ -0,0 +1,106 @@
...
+        return elements;
+    }
+
+    /**
+     * Returns a {@link Stream} over this collection,
+     * whose elements are already cast to {@link TaskUpdate}.
+     *
+     * @return A {@link Stream} as described above.
+     */
+    public Stream<TaskUpdate> streamAsTaskUpdate() {
```
jeremyjams, line 90: streamOfTaskUpdate()
?

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: That seems to be the default behavior - with default value of 60 seconds. However, it is unused in our implementation: we never exceed the core pool size as the maximum pool size is equal to the core and the queue is unbounded.

As far as I understand the `ThreadPoolExecutor`, it works as follows:
- always have `corePoolSize` active threads, even if they do nothing;
- if there is too much work for the `corePoolSize` threads - i.e. the queue is full -, new threads are popped until `maximumPoolSize` threads are reached;
- if there has been too much work at a point and there's now not that much work, some threads are killed to free memory/processors after a period - default to 60 seconds.
It's an automatically scalability process, useful for lots of case but not there :)

src/main/java/com/iexec/core/task/update/TaskUpdatePriorityBlockingQueue.java
mcornaton: I don't really like `of`. The idea here is to tell that the elements in the queue are streamed as `TaskUpdate`s. I don't think `streamOfTaskUpdate` carries the same meaning: it tells this method returns a stream of `TaskUpdate`s but does not say anything about where these objects come from.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -63,14 +79,23 @@
...
             }
-            boolean isOffered = queue.offer(chainTaskId);
+            final Optional<Task> task = taskService.getTaskByChainTaskId(chainTaskId);
+            if (task.isEmpty()) {
+                log.warn("No such task. [chainTaskId: {}]", chainTaskId);
+                return false;
+            }
+
+            taskUpdateExecutor.execute(new TaskUpdate(task.get(), locks, consumer));
             log.info("Published task update request [chainTaskId:{}, queueSize:{}]", chainTaskId, queue.size());
```
jeremyjams, line 93: If we want to keep that log, we could improve them by adding taskUpdate status & deadline

mcornaton: Sure, that could be handy. What about making that log `debug` instead of `info`?

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
jeremyjams: Yes, from my understanding we could set an explicit 0 (https://github.com/JetBrains/jdk8u_jdk/blob/94318f9185757cc33d2b8d527d36be26ac6b7582/src/share/classes/java/util/concurrent/Executors.java#L88).

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
jeremyjams: Yes

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: You're right. Fixed in https://github.com/iExecBlockchainComputing/iexec-core/pull/479/commits/a25e077b0cd01d8a0da68dce2b5f96a6c046328f.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -63,14 +79,26 @@
...
+            if (oTask.isEmpty()) {
+                log.warn("No such task. [chainTaskId: {}]", chainTaskId);
+                return false;
+            }
+
+            final Task task = oTask.get();
+            taskUpdateExecutor.execute(new TaskUpdate(task, locks, consumer));
+            log.debug("Published task update request" +
+                    " [chainTaskId:{}, currentStatus:{}, contributionDeadline:{}, queueSize:{}]",
+                    chainTaskId, task.getChainTaskId(), task.getContributionDeadline(), queue.size());
```
jeremyjams, line 96: Thanks

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
```diff
@@ -105,4 +147,97 @@ public void shouldNotUpdateAtTheSameTime() throws NoSuchFieldException, IllegalA
...
+    }
+
+    @Test
+    void shouldGetInOrderForStatusAndContributionDeadline() throws InterruptedException {
+        final TaskUpdatePriorityBlockingQueue queue = taskUpdateRequestManager.queue;
+
+        final Date d1 = new GregorianCalendar(2021, Calendar.JANUARY, 1).getTime();
+        final Date d2 = new GregorianCalendar(2021, Calendar.JANUARY, 2).getTime();
+
+        TaskUpdate t1 = new TaskUpdate(Task.builder().currentStatus(TaskStatus.RUNNING).contributionDeadline(d1).build(), null, null);
```
jeremyjams, line 224: `buildTaskUpdate(TaskStatus status, Date contributionDeadline)` in yours tests would be cleaner

jeremyjams (approved):
Well-done :)

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
mcornaton: Right. But we have some `TaskUpdate` with `chainTaskId`, `locks` or `consumer` so we need to include them as well to our new method. That's done in https://github.com/iExecBlockchainComputing/iexec-core/pull/479/commits/3c29ec47e8ecb2ae0165d3c1927ef199f76c5d1b.

src/main/java/com/iexec/core/task/update/TaskUpdatePriorityBlockingQueue.java
```diff
@@ -0,0 +1,106 @@
...
+import com.iexec.core.task.Task;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.stream.Stream;
+
+public class TaskUpdatePriorityBlockingQueue extends PriorityBlockingQueue<Runnable> {
```
jbern0rd, line 28: Was it not possible to use `PriorityBlockingQueue<TaskUpdate>` with `TaskUpdate` extending `Runnable` ?
This would allow to reuse all the methods from PriorityBlockingQueue like `contains`, `stream` which would return a `Stream<TaskUpdate>` in our scope.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -44,15 +45,30 @@
...
+    // as new threads are popped only if the queue is full
+    // - which never happens with an unbounded queue.
+    final ThreadPoolExecutor taskUpdateExecutor = new ThreadPoolExecutor(
+            TASK_UPDATE_THREADS_POOL_SIZE,
+            TASK_UPDATE_THREADS_POOL_SIZE,
+            0,
+            TimeUnit.MILLISECONDS,
+            queue
+    );
     private TaskUpdateRequestConsumer consumer;
```
jbern0rd, line 64: I don't really like the kind of circular dependency between `TaskUpdateRequestManager` and `TaskUpdateManager` which implements `TaskUpdateRequestConsumer` and is the bean stored in `consumer` I see.
They seem to be fundamentally intertwined.

src/main/java/com/iexec/core/task/update/TaskUpdatePriorityBlockingQueue.java
jbern0rd: The only method without the `@Override` annotation is only used in tests.

src/main/java/com/iexec/core/task/update/TaskUpdate.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Updates a task.
+     * <br>
+     * 2 updates can be run in parallel if they don't target the same task.
+     * Otherwise, the second update will wait until the first one is achieved.
+     */
+    @Override
+    public void run() {
```
jbern0rd, line 65: As the `Runnable` interface is directly implemented, the `@Override` annotation is unnecessary here.
We implement the method and don't rely on any previous implementation.

src/main/java/com/iexec/core/task/update/TaskUpdate.java
```diff
@@ -0,0 +1,96 @@
...
+
+        String chainTaskId = task.getChainTaskId();
+        log.debug("Selected task [chainTaskId: {}, status: {}]", chainTaskId, task.getCurrentStatus());
+        synchronized (locks.computeIfAbsent(chainTaskId, key -> new Object())) { // require one update on a same task at a time
+            consumer.onTaskUpdateRequest(chainTaskId); // synchronously update task
+        }
+    }
+
+    @Override
+    public int compareTo(TaskUpdate otherTaskUpdate) {
```
jbern0rd, line 78: This or `null` check
```suggestion
    public int compareTo(@NotNull TaskUpdate otherTaskUpdate) {
```

src/main/java/com/iexec/core/task/update/TaskUpdatePriorityBlockingQueue.java
mcornaton: Sadly, that's not how templates work in Java. A `PriorityBlockingQueue<? extends Runnable>` is not compatible with a `PriorityBlockingQueue<Runnable>`.

Regarding the method you're talking about, there are 3 non-annotated methods: `takeAll`, `streamAsTaskUpdate` and `containsTask`. I think you're talking about `takeAll`, right? I thought about putting it into the test class but I was not sure about this design. Wouldn't that method be handy someday?

src/main/java/com/iexec/core/task/update/TaskUpdate.java
mcornaton: I don't get your point. `@Override` tells the compiler to check the method does effectively override or implement a method declared in a supertype. That's some security that should, IMHO, always be present.

src/main/java/com/iexec/core/task/update/TaskUpdate.java
mcornaton: I agree for `null` check. `@NotNull` is an annotation provided by JetBrains and I'd prefer not to add some IDE-dependent dependency.

jbern0rd: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Comparator.html#compare(T,T)
Can throw `NullPointerException`, we just need to handle it the day we use `compareTo` (does not seem to be the case right now).

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: Sure, I find it strange, too. I wonder whether `TaskUpdateManager#publishUpdateTaskRequest` should be into `TaskUpdateRequestManager`. BTW, why not using the already existing `TaskUpdateRequestManager#publishRequest` method?
@jeremyjams or @zguesmi, do you have any clue on why it has been designed this way?

src/main/java/com/iexec/core/task/update/TaskUpdate.java
mcornaton: It's used inside `TaskUpdatePriorityBlockingQueue` as the natural-ordering comparator.
Having a look at your link, throwing an NPE seems to be the thing to do:
```
compare(x, y) must throw an exception if and only if compare(y, x) throws an exception.
```
Let me update this!

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
jeremyjams: The initial idea was to expose a single `TaskUpdateManager` entrypoint to others services (`TaskUpdateManager#publishUpdateTaskRequest`). Internal updates would be handle by a dedicated class (`TaskUpdateRequestManager`) in order to avoid overloading `TaskUpdateManager` more and make testing easier.

What do you propose?

src/main/java/com/iexec/core/task/update/TaskUpdate.java
jbern0rd: If you implement an interface without defining implementations for its methods, the compilation will fail.
I don't see the added security of the `@Override` annotation in this case, unless it helps to quickly detect inherited or implemented methods in the `TaskUpdate` class.

mcornaton: I see it the other way: it enforces the class to implement the `Runnable` interface. It's handy if we ever refactor it and miss this `implements Runnable` declaration.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: I see the point.
As of now, there's only a few public methods in `TaskUpdateManager`:
- `onTaskUpdateRequest`, called from `TaskUpdateRequestManager`,
- `publishUpdateTaskRequest`, calling `TaskUpdateRequesterManager#publishRequest`;
- `isConsensusReached`, called from `ReplicateSupplyService` and `TaskUpdateManager` itself.

Regarding these 3 methods, I don't think `TaskUpdateManager` is an entrypoint anymore.
What about the following:
- `TaskUpdateRequestManager` becomes this entrypoint and directly calls `TaskUpdateManager#updateTask`;
- methods that currently call `TaskUpdateManager#publishUpdateTaskRequest` call `TaskUpdateRequestManager#publishRequest` instead;
- `isConsensusReached` is moved to `TaskService`: almost all classes depending on `TaskUpdateManager` also depend on `TaskService` so that does not look to be a big deal;
- `onTaskUpdateRequest` and `publishUpdateTaskRequest` methods are deleted, and so is the interface `TaskUpdateRequestConsumer`.

jeremyjams: Yes, it's fine to me. Just make sure previous points are addressed before refactoring

mcornaton: I'd like to work on it once this PR is merged. That would probably be cleaner :)

src/main/java/com/iexec/core/task/update/TaskUpdate.java
jbern0rd: I understand your point, thanks

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
jbern0rd: Seems a good idea to me, both on the architecture side and on implementing this in another PR

src/main/java/com/iexec/core/task/update/TaskUpdatePriorityBlockingQueue.java
jbern0rd: My main point was on `PriorityBlockingQueue<TaskUpdate>`.
If it's not possible to use this, my second comment becomes irrelevant and we can keep the new methods. :+1:

jbern0rd (approved):
Thanks for this :+1:

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
zguesmi: +1 to what you're suggesting @mcornaton .

mcornaton: Suggestion implemented in https://github.com/iExecBlockchainComputing/iexec-core/pull/484.

PR #480, opened 2021-12-17
Bugfix/too many contribution deadline detector calls

jbern0rd:
Fix a call to ContributionTimeoutTaskDetector done for every task detected.
A single call is sufficient.

### DIFF
--- a/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
@@ -239,6 +239,7 @@ public class ReplicateSupplyService {
                 break;
             case PLEASE_ABORT:
                 taskNotificationExtra.setTaskAbortCause(getTaskAbortCause(task));
+                break;
             default:
                 break;
         }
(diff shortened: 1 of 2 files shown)


mcornaton (approved):
Thanks!

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -112,15 +112,9 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
-                    if (task.isContributionDeadlineReached()) {
-                        contributionTimeoutTaskDetector.detect();
-                        return false;
-                    } else {
-                        return true;
-                    }
-                })
+                .filter(task -> ! task.isContributionDeadlineReached())
                 .collect(Collectors.toCollection(ArrayList::new));
+        if (validTasks.size() != runningTasks.size()) contributionTimeoutTaskDetector.detect();
```
zguesmi, line 117: It was agreed that we don't use one-liners for readability purposes.

zguesmi (approved):
Thank for cleaning that.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jbern0rd: fixed, new commit pushed

PR #483, opened 2021-12-23
Bugfix/task not active

jbern0rd:
* Add check on task current status when iterating over valid tasks
* Add tests to cover new check and missing cases
* Rework verification points in check methods

### DIFF
--- a/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
@@ -191,7 +191,7 @@ public class ReplicateSupplyService {
             String chainTaskId = task.getChainTaskId();
 
             Optional<Replicate> oReplicate = replicatesService.getReplicate(chainTaskId, walletAddress);
-            if (!oReplicate.isPresent()) {
+            if (oReplicate.isEmpty()) {
                 continue;
             }
             Replicate replicate = oReplicate.get();
@@ -204,7 +204,7 @@ public class ReplicateSupplyService {
                 continue;
             }
             Optional<TaskNotificationType> taskNotificationType = getTaskNotificationType(task, replicate, blockNumber);
-            if (!taskNotificationType.isPresent()) {
+            if (taskNotificationType.isEmpty()) {
                 continue;
             }
             TaskNotificationExtra taskNotificationExtra =
@@ -289,7 +289,9 @@ public class ReplicateSupplyService {
         String chainTaskId = task.getChainTaskId();
         String walletAddress = replicate.getWalletAddress();
 
-        if (!replicate.getLastRelevantStatus().isPresent()) return Optional.empty();
+        if (replicate.getLastRelevantStatus().isEmpty()) {
+            return Optional.empty();
+        }
 
         boolean beforeContributing = replicate.isBeforeStatus(ReplicateStatus.CONTRIBUTING);
         boolean didReplicateStartContributing = replicate.getLastRelevantStatus().get().equals(ReplicateStatus.CONTRIBUTING);
@@ -310,10 +312,14 @@ public class ReplicateSupplyService {
 
         // we read the replicate from db to consider the changes added in the previous case
         Optional<Replicate> oReplicateWithLatestChanges = replicatesService.getReplicate(chainTaskId, walletAddress);
-        if (!oReplicateWithLatestChanges.isPresent()) return Optional.empty();
+        if (oReplicateWithLatestChanges.isEmpty()) {
+            return Optional.empty();
+        }
 
         Replicate replicateWithLatestChanges = oReplicateWithLatestChanges.get();
-        if (!replicateWithLatestChanges.getLastRelevantStatus().isPresent()) return Optional.empty();
+        if (replicateWithLatestChanges.getLastRelevantStatus().isEmpty()) {
+            return Optional.empty();
+        }
 
         boolean didReplicateContribute = replicateWithLatestChanges.getLastRelevantStatus().get()
                 .equals(ReplicateStatus.CONTRIBUTED);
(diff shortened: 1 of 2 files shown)


mcornaton (approved):
LGTM! Thanks :)

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -118,6 +118,8 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
             contributionTimeoutTaskDetector.detect();
```
jeremyjams, line 118: This method could take a lot of time, increasing chances to have an task status from the beginning until the end. It should be async in order to delegate responsibility outside.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -118,6 +118,8 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
         }
 
+        // TODO : Remove this, the optional can never be empty
+        // This is covered in workerService.canAcceptMoreWorks
```
jeremyjams, line 122: #getWorker should be done before #getInitializedOrRunningTasks to avoid losing time. I would add most safety checks in  previous task stream filters (before this for or directly in the database query).

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -127,14 +129,21 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
         for (Task task : validTasks) {
             String chainTaskId = task.getChainTaskId();
 
-            // no need to ge further if the consensus is already reached on-chain
+            // Check if task is still in contribution phase with INITIALIZED or RUNNING status
+            Optional<Task> upToDateTask = taskService.getTaskByChainTaskId(chainTaskId);
+            if (upToDateTask.isEmpty() || ! TaskStatus.isInContributionPhase(upToDateTask.get().getCurrentStatus())) {
```
jeremyjams, line 134: I see your point but that patch looks a bit strange to me.
The design of this method should be updated.

In the mean time, in addition to moving most safety checks outside this for loop, I would move your patch right before `replicatesService.addNewReplicate(chainTaskId, walletAddress);`, in order to decrease chances of giving a task/replicate with an outdated status.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jbern0rd: Are you sure ? This method will:
* Fetch all tasks `INITIALIZED` or `RUNNING` from the database
* Call `TaskUpdateManager#publishUpdateTaskRequest` which call `TaskUpdateRequestManager#publishRequest`
The latter is already async.
But it can be made async to avoid impact from refactoring on `TaskUpdateManager` or `TaskUpdateRequestManager`.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jbern0rd: In fact, this call only serve to retrieve worker instance for `worker.isTeeEnabled()` check.
If we want to go further right now, I would prefer to access TEE worker status through a WorkerService call.
It may be best to do it in another PR to avoid mixing 2 subjects. That's why I added TODO on lines 121 and 146.

I agree that I would prefer to have:
`boolean isWorkerTeeEnabled = workerService.isTeeEnabled(walletAddress)`
outside from the loop to only compute the status once

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Don't you agree `contributionTimeoutTaskDetector.detect()` could take quite some time before returning?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jbern0rd: We need to discuss on how this method should be updated then to specify and implement things correctly.

I don't know if we want to do it right now or send this first patch and rework this implementation early January.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: The `fetch` task could take some time, sure. By the way, if another worker asks for a new replicate to compute at the same time, it'll trigger another `contributionTimeoutTaskDetector.detect()` call. We should add some mechanism to avoid that scenario.
(Some way to prevent `contributionTimeoutTaskDetector.detect()` from starting twice in parallel?)

jbern0rd: For me, there is a problem with the current implementation which mixes:
* Scheduled async executions in `ContributionTimeoutTaskDetector`
* Direct calls to `ContributionTimeoutTaskDetector#detect`
I think the whole thing should be refactored and I would like to create and fire `ContributionTimeoutTaskEvent` instances and handle those events with a listener in `ContributionTimeoutTaskDetector`.
See https://www.baeldung.com/spring-events

jeremyjams: Yes, instead of triggering the whole detector (which re-read tasks) we should publish the existing `com.iexec.core.task.event.ContributionTimeoutEvent` when we found a `task.isContributionDeadlineReached()` from the filter since we are sure we got one.

jbern0rd: I would like to address this in a next PR if it's okay for everyone.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jbern0rd: Check moved to a later part of the code, just before retrieving replicates and computing if more contributions are required
See 99555cb

jeremyjams (approved):
Alright, let's start with that patch and increment/re-design in next PRs :)

jbern0rd:
Thanks, let's do it this way !

PR #484, opened 2021-12-23
Refactor task update

mcornaton:
Refactor task update part, according to what has been discussed in https://github.com/iExecBlockchainComputing/iexec-core/pull/479#discussion_r774485124.


src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -24,7 +24,7 @@
 import java.util.List;
 import java.util.Optional;
 
-interface TaskRepository extends MongoRepository<Task, String> {
+public interface TaskRepository extends MongoRepository<Task, String> {
```
jeremyjams, line 27: Why public?

src/main/java/com/iexec/core/task/update/TaskUpdate.java
```diff
@@ -69,9 +66,7 @@ public void run() {
 
         String chainTaskId = task.getChainTaskId();
         log.debug("Selected task [chainTaskId: {}, status: {}]", chainTaskId, task.getCurrentStatus());
-        synchronized (locks.computeIfAbsent(chainTaskId, key -> new Object())) { // require one update on a same task at a time
-            consumer.onTaskUpdateRequest(chainTaskId); // synchronously update task
-        }
+        taskUpdater.accept(chainTaskId);
```
jeremyjams, line 69: :+1:

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: It is used by `TaskUpdateManager`, which I moved into the `com.iexec.core.task.update` package - where it should have been. A package-private visibility is insufficient, then.

jbern0rd: Shouldn't `TaskRepository` be accessed through `TaskService` in this case ?

jbern0rd: This could allow to remove TaskRepository as a dependency from TaskUpdateManager, which can be a good idea because there are lots of injected dependency, which is not recommended.
See code quality reports and **Methods should not have too many parameters** rule.

mcornaton: You're right. I'll work on that.

mcornaton: Done in 100df2d.

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -93,6 +103,22 @@ public TaskService(TaskRepository taskRepository) {
...
+     * @return An {@link Optional<Task>} if task exists, {@link Optional#empty()} otherwise.
+     */
+    public Optional<Task> updateTask(Task task) {
+        final Optional<Task> knownTask = taskRepository
+                .findByChainTaskId(task.getChainTaskId());
+        if (knownTask.isPresent()) {
+            return Optional.of(taskRepository.save(task));
+        } else {
+            return Optional.empty();
+        }
```
zguesmi, line 119:
```suggestion
        return taskRepository
                .findByChainTaskId(task.getChainTaskId())
                .map(existingTask -> taskRepository.save(task));
```
or
```suggestion
        return taskRepository.existsByChainTaskId(task.getChainTaskId())
                ? Optional.of(taskRepository.save(task))
                : Optional.empty();
```

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -100,7 +83,7 @@ void updateTask(String chainTaskId) {
             // Eventually should fire a "final deadline reached" notification to worker,
             // but here let's just trigger an updateTask() leading to a failed status
             // which will itself fire a generic "abort" notification
-            publishUpdateTaskRequest(chainTaskId);
+            toFailed(task);
```
zguesmi, line 86: Could you update comments please;

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -180,9 +163,14 @@ Task updateTaskStatusAndSave(Task task, TaskStatus newStatus) {
...
-        Task savedTask = taskRepository.save(task);
-        log.info("UpdateTaskStatus suceeded [chainTaskId:{}, currentStatus:{}, newStatus:{}]", task.getChainTaskId(), currentStatus, newStatus);
-        return savedTask;
+        Optional<Task> savedTask = taskService.updateTask(task);
+        if (savedTask.isPresent()) {
+            log.info("UpdateTaskStatus succeeded [chainTaskId:{}, currentStatus:{}, newStatus:{}]", task.getChainTaskId(), currentStatus, newStatus);
+            return savedTask.get();
+        } else {
+            log.warn("UpdateTaskStatus failed [chainTaskId:{}, currentStatus:{}, wishedStatus:{}]", task.getChainTaskId(), currentStatus, newStatus);
+            return null;
```
zguesmi, line 172: Does the calling methods verify null values?

src/main/java/com/iexec/core/task/TaskService.java
mcornaton: With a `.map` and a method reference, it's much cleaner. Thanks!

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
mcornaton: It is never checked as return value is never used.

src/main/java/com/iexec/core/task/TaskService.java
mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-core/pull/484/commits/3df6dcf53f4735b8dbaecea4f108532047c1d217.

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
mcornaton: Done in e0a4e54.

src/main/java/com/iexec/core/task/TaskService.java
zguesmi: Thanks. But with the method reference, the task fetched from db will be saved, not the one passed as argument).

mcornaton: Right, I had misread your comment. Thanks!
Fixed in https://github.com/iExecBlockchainComputing/iexec-core/pull/484/commits/01e1299635cd12945bdd1be95e0c537e68df68f3.

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
jbern0rd, line 172: This value is never checked in calling methods.
If the update fails, there will only be a warn in the log and the scheduler will never be aware of it.
What should the behavior be when calling `updateTaskStatusAndSave` in succession if one of them fails ?
example
```
updateTaskStatusAndSave(task, FINALIZE_FAILED);
updateTaskStatusAndSave(task, FAILED);
```

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -174,4 +196,18 @@ private void setTaskAccessForNewReplicateLock(String chainTaskId, boolean isTask
         taskAccessForNewReplicateLock.replace(chainTaskId, isTaskBeingAccessedForNewReplicate);
     }
 
+    public boolean isConsensusReached(Task task) {
```
jbern0rd, line 199:
```suggestion
    public boolean isConsensusReached(String chainTaskId) {
```
What do you think ?

src/main/java/com/iexec/core/task/TaskStatusChange.java
```diff
@@ -17,6 +17,7 @@
 package com.iexec.core.task;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.iexec.core.task.TaskStatus;
```
jbern0rd, line 20: Unnecessary import ?

src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
```diff
@@ -106,7 +106,7 @@ void shouldNotGetAnyReplicateSinceWorkerDoesNotExist() {
         Optional<WorkerpoolAuthorization> oAuthorization =
                 replicateSupplyService.getAuthOfAvailableReplicate(workerLastBlock, WALLET_WORKER_1);
         assertThat(oAuthorization).isEmpty();
-        Mockito.verifyNoInteractions(web3jService, taskService, contributionTimeoutTaskDetector, taskUpdateManager, replicatesService, signatureService, smsService);
+        Mockito.verifyNoInteractions(web3jService, taskService, contributionTimeoutTaskDetector, replicatesService, signatureService, smsService);
```
jbern0rd, line 109:
```suggestion
        Mockito.verifyNoInteractions(web3jService, taskUpdateRequestManager, taskService, contributionTimeoutTaskDetector, replicatesService, signatureService, smsService);
```
I think `taskUpdateManager` should have been replaced with `taskUpdateRequestManager` in some tests.
This is to ensure we never call some objects which mean we did not go on some code paths.
There is a loss of precision with the last fix imho

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
```diff
@@ -91,22 +94,25 @@ void shouldNotUpdateAtTheSameTime() {
...
                 .build();
+        final Consumer<String> taskUpdater = chainTaskId -> {
+            synchronized (locks.computeIfAbsent(chainTaskId, key -> new Object())) {
+                final int updateId = (int) System.nanoTime() % Integer.MAX_VALUE;
+                taskForUpdateId.put(updateId, chainTaskId);
+                for (int i = 0; i < callsPerUpdate; i++) {
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(random.nextInt(10));
+                    } catch (InterruptedException ignored) {
+                    }
```
jbern0rd, line 108: Isn't this dangerous, even if we are in tests ?

mcornaton: I don't think so. This exception is only thrown when the thread is interrupted. That means the test won't complete, so it'll be considered as failed.
We could add some log if you think it is useful, I don't have any opinion on that.

src/main/java/com/iexec/core/task/TaskStatusChange.java
mcornaton: Right, thanks!

src/main/java/com/iexec/core/task/TaskService.java
mcornaton: Well, yes, that's probably better. I'll update that!

src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
mcornaton: Right, updated in 8b05a75.

src/test/java/com/iexec/core/task/update/TaskUpdateRequestManagerTests.java
jbern0rd: Maybe just to be safe and be sure the test will fail and stop fast enough
```
} catch (InterruptedException e) {
    throw new RuntimeException(e);
}
```
https://stackoverflow.com/questions/60619543/how-to-throw-an-exception-in-a-consumer-java-8

mcornaton: Ok, let's do this!

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
mcornaton: Not sure whether it's a good idea as it could lead to infinite loop if `taskService.updateTask(...)` always fails - which would be a serious issue itself.

BTW, I've just checked the `repository#save` method JavaDoc and it states the following:
> @return the saved entity; will never be {@literal null}.

So we should never encounter this case if the task exists. To be clear, I'll add a more explicit comment to the `updateTaskStatusAndSave` method.

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -90,7 +92,7 @@ public TaskUpdateRequestManager(TaskService taskService) {
             }
 
             final Task task = oTask.get();
-            taskUpdateExecutor.execute(new TaskUpdate(task, locks, consumer));
+            taskUpdateExecutor.execute(new TaskUpdate(task, this::updateTask));
```
jeremyjams, line 95: :dart:

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
```diff
@@ -102,12 +104,9 @@ public TaskUpdateRequestManager(TaskService taskService) {
...
     }
 
-    /**
-     * Authorize one TaskUpdateRequest consumer subscription at a time.
-     * @param consumer
-     * @return
-     */
-    public void setRequestConsumer(final TaskUpdateRequestConsumer consumer) {
-        this.consumer = consumer;
+    private void updateTask(String chainTaskId) {
```
jeremyjams, line 107: Maybe add a todo note for using `ContextuaLockRunner`

jeremyjams (approved):

src/main/java/com/iexec/core/task/update/TaskUpdateRequestManager.java
mcornaton: Good idea! See https://github.com/iExecBlockchainComputing/iexec-core/pull/484/commits/4b0847fb0d03a45df49dd33d66961a3550ea210c.

jbern0rd (approved):

PR #490, opened 2022-01-10
Add a fix for `TASK_NOT_ACTIVE`

mcornaton:
Replicates list should be entirely fixed in `ReplicateSupplyService#getAuthOfAvailableReplicate`.

### DIFF
--- a/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
@@ -128,10 +128,11 @@ public class ReplicateSupplyService {
 
         for (Task task : validTasks) {
             String chainTaskId = task.getChainTaskId();
+            final List<Replicate> replicates = replicatesService.getReplicates(chainTaskId);
 
             // no need to go further if the consensus is already reached on-chain
             // the task should be updated since the consensus is reached but it is still in RUNNING status
-            if (taskUpdateManager.isConsensusReached(task)) {
+            if (taskUpdateManager.isConsensusReached(chainTaskId, replicates)) {
                 taskUpdateManager.publishUpdateTaskRequest(chainTaskId);
                 continue;
             }
@@ -162,7 +163,6 @@ public class ReplicateSupplyService {
                 continue;
             }
 
-            final List<Replicate> replicates = replicatesService.getReplicates(chainTaskId);
             final boolean taskNeedsMoreContributions = ConsensusHelper.doesTaskNeedMoreContributionsForConsensus(
                     chainTaskId,
                     replicates,
@@ -339,8 +339,8 @@ public class ReplicateSupplyService {
                 .equals(ReplicateStatus.CONTRIBUTED);
 
         if (didReplicateContribute) {
-
-            if (!taskUpdateManager.isConsensusReached(task)) {
+            final List<Replicate> replicates = replicatesService.getReplicates(chainTaskId);
+            if (!taskUpdateManager.isConsensusReached(chainTaskId, replicates)) {
                 return Optional.of(TaskNotificationType.PLEASE_WAIT);
             }
 
--- a/src/main/java/com/iexec/core/replicate/ReplicatesService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicatesService.java
@@ -168,9 +168,9 @@ public class ReplicatesService {
         return addressReplicates.size();
     }
 
-    public int getNbValidContributedWinners(String chainTaskId, String contributionHash) {
+    public int getNbValidContributedWinners(List<Replicate> replicates, String contributionHash) {
         int nbValidWinners = 0;
-        for (Replicate replicate : getReplicates(chainTaskId)) {
+        for (Replicate replicate : replicates) {
             Optional<ReplicateStatus> oStatus = replicate.getLastRelevantStatus();
             if (oStatus.isPresent() && oStatus.get().equals(CONTRIBUTED)
                     && contributionHash.equals(replicate.getContributionHash())) {
--- a/src/main/java/com/iexec/core/task/TaskUpdateManager.java
+++ b/src/main/java/com/iexec/core/task/TaskUpdateManager.java
@@ -278,9 +278,9 @@ public class TaskUpdateManager implements TaskUpdateRequestConsumer  {
         }
     }
 
-    public boolean isConsensusReached(Task task) {
+    public boolean isConsensusReached(String chainTaskId, List<Replicate> replicates) {
 
-        Optional<ChainTask> optional = iexecHubService.getChainTask(task.getChainTaskId());
+        Optional<ChainTask> optional = iexecHubService.getChainTask(chainTaskId);
         if (optional.isEmpty()) return false;
 
         ChainTask chainTask = optional.get();
@@ -288,7 +288,7 @@ public class TaskUpdateManager implements TaskUpdateRequestConsumer  {
         boolean isChainTaskRevealing = chainTask.getStatus().equals(ChainTaskStatus.REVEALING);
 
         int onChainWinners = chainTask.getWinnerCounter();
-        int offChainWinners = isChainTaskRevealing ? replicatesService.getNbValidContributedWinners(task.getChainTaskId(), chainTask.getConsensusValue()) : 0;
+        int offChainWinners = isChainTaskRevealing ? replicatesService.getNbValidContributedWinners(replicates, chainTask.getConsensusValue()) : 0;
         boolean offChainWinnersGreaterOrEqualsOnChainWinners = offChainWinners >= onChainWinners;
 
         return isChainTaskRevealing && offChainWinnersGreaterOrEqualsOnChainWinners;
@@ -296,22 +296,24 @@ public class TaskUpdateManager implements TaskUpdateRequestConsumer  {
 
     void running2ConsensusReached(Task task) {
         boolean isTaskInRunningStatus = task.getCurrentStatus().equals(RUNNING);
-        boolean isConsensusReached = isConsensusReached(task);
+        final String chainTaskId = task.getChainTaskId();
+        List<Replicate> replicates = replicatesService.getReplicates(chainTaskId);
+        boolean isConsensusReached = isConsensusReached(chainTaskId, replicates);
 
         if (isTaskInRunningStatus && isConsensusReached) {
-            Optional<ChainTask> optional = iexecHubService.getChainTask(task.getChainTaskId());
+            Optional<ChainTask> optional = iexecHubService.getChainTask(chainTaskId);
             if (optional.isEmpty()) return;
             ChainTask chainTask = optional.get();
 
             // change the the revealDeadline and consensus of the task from the chainTask info
             task.setRevealDeadline(new Date(chainTask.getRevealDeadline()));
             task.setConsensus(chainTask.getConsensusValue());
-            long consensusBlockNumber = iexecHubService.getConsensusBlock(task.getChainTaskId(), task.getInitializationBlockNumber()).getBlockNumber();
+            long consensusBlockNumber = iexecHubService.getConsensusBlock(chainTaskId, task.getInitializationBlockNumber()).getBlockNumber();
             task.setConsensusReachedBlockNumber(consensusBlockNumber);
             updateTaskStatusAndSave(task, CONSENSUS_REACHED);
 
             applicationEventPublisher.publishEvent(ConsensusReachedEvent.builder()
-                    .chainTaskId(task.getChainTaskId())
+                    .chainTaskId(chainTaskId)
                     .consensus(task.getConsensus())
                     .blockNumber(task.getConsensusReachedBlockNumber())
                     .build());
--- a/src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
+++ b/src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
@@ -203,6 +203,7 @@ class ReplicateSupplyServiceTests {
                 .teeEnabled(false)
                 .lastAliveDate(new Date())
                 .build();
+        List<Replicate> replicates = List.of(new Replicate());
 
         Task runningTask = new Task(DAPP_NAME, COMMAND_LINE, 5, CHAIN_TASK_ID);
         runningTask.setMaxExecutionTime(maxExecutionTime);
@@ -215,13 +216,14 @@ class ReplicateSupplyServiceTests {
                 .thenReturn(Collections.singletonList(runningTask));
         when(workerService.getWorker(WALLET_WORKER_2)).thenReturn(Optional.of(worker));
         when(taskService.getTaskByChainTaskId(CHAIN_TASK_ID)).thenReturn(Optional.of(runningTask));
-        when(taskUpdateManager.isConsensusReached(runningTask)).thenReturn(true);
+        when(replicatesService.getReplicates(CHAIN_TASK_ID)).thenReturn(replicates);
+        when(taskUpdateManager.isConsensusReached(CHAIN_TASK_ID, replicates)).thenReturn(true);
 
         Optional<WorkerpoolAuthorization> oAuthorization =
                 replicateSupplyService.getAuthOfAvailableReplicate(workerLastBlock, WALLET_WORKER_2);
         assertThat(oAuthorization).isEmpty();
-        Mockito.verify(taskUpdateManager).isConsensusReached(runningTask);
-        Mockito.verifyNoInteractions(replicatesService, signatureService, smsService);
+        Mockito.verify(taskUpdateManager).isConsensusReached(CHAIN_TASK_ID, replicates);
+        Mockito.verifyNoInteractions(signatureService, smsService);
         assertTaskAccessForNewReplicateLockNeverUsed();
     }
 
@@ -393,8 +395,8 @@ class ReplicateSupplyServiceTests {
 
         // the call should only happen once over the two tasks
         Mockito.verify(contributionTimeoutTaskDetector).detect();
-        Mockito.verify(taskUpdateManager).isConsensusReached(task1);
-        Mockito.verify(taskUpdateManager, Mockito.never()).isConsensusReached(taskDeadlineReached);
+        Mockito.verify(taskUpdateManager).isConsensusReached(CHAIN_TASK_ID, List.of(replicate));
+        Mockito.verify(taskUpdateManager, Mockito.never()).isConsensusReached(CHAIN_TASK_ID_2, List.of(replicate));
     }
 
     @Test
@@ -425,7 +427,7 @@ class ReplicateSupplyServiceTests {
 
         assertThat(oAuthorization).isEmpty();
         Mockito.verify(taskService).isTaskBeingAccessedForNewReplicate(CHAIN_TASK_ID);
-        Mockito.verifyNoInteractions(contributionTimeoutTaskDetector, replicatesService, signatureService, smsService);
+        Mockito.verifyNoInteractions(contributionTimeoutTaskDetector, signatureService, smsService);
         assertTaskAccessForNewReplicateLockNeverUsed();
     }
 
@@ -531,7 +533,7 @@ class ReplicateSupplyServiceTests {
 
         assertThat(oAuthorization).isEmpty();
         Mockito.verify(taskService, Mockito.never()).isTaskBeingAccessedForNewReplicate(CHAIN_TASK_ID);
-        Mockito.verifyNoInteractions(contributionTimeoutTaskDetector, replicatesService, signatureService, smsService);
+        Mockito.verifyNoInteractions(contributionTimeoutTaskDetector, signatureService, smsService);
         assertTaskAccessForNewReplicateLockNeverUsed();
     }
 
@@ -682,6 +684,7 @@ class ReplicateSupplyServiceTests {
         List<Task> taskList = getStubTaskList(TaskStatus.RUNNING);
         Optional<Replicate> replicate1 = getStubReplicate(ReplicateStatus.CONTRIBUTING);
         Optional<Replicate> replicate2 = getStubReplicate(ReplicateStatus.CONTRIBUTED);
+        final List<Replicate> replicates = List.of(replicate1.get(), replicate2.get());
 
         when(workerService.getChainTaskIds(WALLET_WORKER_1)).thenReturn(ids);
         when(taskService.getTasksByChainTaskIds(ids)).thenReturn(taskList);
@@ -692,7 +695,8 @@ class ReplicateSupplyServiceTests {
                 .thenReturn(getStubAuth());
         when(replicatesService.didReplicateContributeOnchain(CHAIN_TASK_ID, WALLET_WORKER_1))
                 .thenReturn(true);
-        when(taskUpdateManager.isConsensusReached(taskList.get(0))).thenReturn(false);
+        when(replicatesService.getReplicates(CHAIN_TASK_ID)).thenReturn(replicates);
+        when(taskUpdateManager.isConsensusReached(CHAIN_TASK_ID, replicates)).thenReturn(false);
 
         List<TaskNotification> missedTaskNotifications =
                 replicateSupplyService.getMissedTaskNotifications(blockNumber, WALLET_WORKER_1);
@@ -718,6 +722,7 @@ class ReplicateSupplyServiceTests {
         List<Task> taskList = getStubTaskList(TaskStatus.RUNNING);
         Optional<Replicate> replicate1 = getStubReplicate(ReplicateStatus.CONTRIBUTING);
         Optional<Replicate> replicate2 = getStubReplicate(ReplicateStatus.CONTRIBUTED);
+        List<Replicate> replicates = List.of(replicate1.get(), replicate2.get());
 
         when(workerService.getChainTaskIds(WALLET_WORKER_1)).thenReturn(ids);
         when(taskService.getTasksByChainTaskIds(ids)).thenReturn(taskList);
@@ -728,7 +733,8 @@ class ReplicateSupplyServiceTests {
                 .thenReturn(getStubAuth());
         when(replicatesService.didReplicateContributeOnchain(CHAIN_TASK_ID, WALLET_WORKER_1))
                 .thenReturn(true);
-        when(taskUpdateManager.isConsensusReached(taskList.get(0))).thenReturn(true);
+        when(replicatesService.getReplicates(CHAIN_TASK_ID)).thenReturn(replicates);
+        when(taskUpdateManager.isConsensusReached(CHAIN_TASK_ID, replicates)).thenReturn(true);
 
         List<TaskNotification> missedTaskNotifications =
                 replicateSupplyService.getMissedTaskNotifications(blockNumber, WALLET_WORKER_1);
--- a/src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
+++ b/src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
@@ -381,7 +381,7 @@ public class ReplicateServiceTests {
                 .thenReturn(Optional.of(replicatesList));
 
         assertThat(replicatesService.getNbValidContributedWinners(
-                CHAIN_TASK_ID,
+                replicatesList.getReplicates(),
                 contributionHash
         )).isOne();        
     }


src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -339,8 +339,8 @@ private TaskNotificationExtra getTaskNotificationExtra(Task task, TaskNotificati
                 .equals(ReplicateStatus.CONTRIBUTED);
 
         if (didReplicateContribute) {
-
-            if (!taskUpdateManager.isConsensusReached(task)) {
+            final List<Replicate> replicates = replicatesService.getReplicates(chainTaskId);
+            if (!taskUpdateManager.isConsensusReached(chainTaskId, replicates)) {
```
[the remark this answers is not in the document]
mcornaton: Good idea, thanks! That's fixed in fb2be03.

src/main/java/com/iexec/core/replicate/ReplicatesList.java
```diff
@@ -43,12 +43,12 @@
...
     private List<Replicate> replicates;
 
-    ReplicatesList(String chainTaskId) {
+    public ReplicatesList(String chainTaskId) {
         this.chainTaskId = chainTaskId;
         this.replicates = new ArrayList<>();
     }
 
-    ReplicatesList(String chainTaskId, List<Replicate> replicates) {
+    public ReplicatesList(String chainTaskId, List<Replicate> replicates) {
```
jeremyjams, line 51: That one is useless

mcornaton: Sure, but I find it better to be consistent with the other constructor.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -128,10 +128,17 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
 
         for (Task task : validTasks) {
             String chainTaskId = task.getChainTaskId();
+            final Optional<ReplicatesList> oReplicatesList = replicatesService.getReplicatesList(chainTaskId);
+            // If there's no replicatesList, the task is probably not initialized.
+            if (oReplicatesList.isEmpty()) {
```
zguesmi, line 133: An initialized task will have 0 replicates, because it is not running yet.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -128,10 +128,17 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
 
         for (Task task : validTasks) {
             String chainTaskId = task.getChainTaskId();
+            final Optional<ReplicatesList> oReplicatesList = replicatesService.getReplicatesList(chainTaskId);
+            // If there's no replicatesList, the task is probably not initialized.
```
zguesmi, line 132:
```suggestion
            // If there's no replicatesList, the task is probably not initialized.
            // FYI: initialized but not-yet-running tasks will have a defined ReplicateList with size == 0.
```

zguesmi: This filtering should probably be done in the method `taskService.getInitialiazedOrRunningTasks()`.

zguesmi: We can add a TODO comment to filter those tasks in-db directly if it is possible with mongo.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -168,9 +168,9 @@ public int getNbReplicatesContainingStatus(String chainTaskId, ReplicateStatus..
         return addressReplicates.size();
     }
 
-    public int getNbValidContributedWinners(String chainTaskId, String contributionHash) {
+    public int getNbValidContributedWinners(List<Replicate> replicates, String contributionHash) {
```
zguesmi, line 171: This method has become static.

src/main/java/com/iexec/core/task/TaskUpdateManager.java
```diff
@@ -278,40 +278,49 @@ void initialized2Running(Task task) {
...
         boolean offChainWinnersGreaterOrEqualsOnChainWinners = offChainWinners >= onChainWinners;
 
         return isChainTaskRevealing && offChainWinnersGreaterOrEqualsOnChainWinners;
     }
 
     void running2ConsensusReached(Task task) {
         boolean isTaskInRunningStatus = task.getCurrentStatus().equals(RUNNING);
-        boolean isConsensusReached = isConsensusReached(task);
+        final String chainTaskId = task.getChainTaskId();
+        final Optional<ReplicatesList> oReplicatesList = replicatesService.getReplicatesList(chainTaskId);
```
zguesmi, line 302: Wouldn't this result in the same synchronization problems ? The task and the replicates list hasn't been fetched at the same time. I think the replicateList should be fetched from the db by the caller at the same time as the task and passed to this method.
This makes me think about a `MetaTask` object that contains at the same time, the task details and the replicatesList. That way we guarantee that the task and it's replicate are always synchronized.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Static methods can't be mocked so I'm not sure whether it should be marked as static.

src/main/java/com/iexec/core/task/TaskUpdateManager.java
mcornaton: I don't think there could be the same sync problem here:
- a same task can't be updated twice at the same time so its status won't change between its retrieval and the start of `running2ConsensusReached`;
- replicates can be updated between the task retrieval and the start of `running2ConsensusReached` but I don't think it matters. The "worst" case would be an update of one or multiple replicates in that period, but that would simply lead to another result for the method... Which would have been computed later otherwise.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: I disagree. As @jbern0rd pointed out a few weeks ago, there could be some desynchronization there so we should check the current status of the replicates in the `for`-loop, the nearest to its use.

zguesmi: The method `taskService.getInitialiazedOrRunningTasks()` returns non-initialized non-running tasks. Do you find that coherent?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
zguesmi: Ok we can keep it non-static if it's more convenient. Could you kindly add a comment to mention that for the future?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: When they are retrieved, they are initialized or running. But their status can change between their retrieval and their use.
However, this filtering should never be used as `replicatesList` should not be back to an empty list. That's just a safeguard.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Sure, see https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/d3208756bc211de878afc6f5075ed62f8561ce7f.
By the way, I'm wondering whether we should keep both version of `getNbValidContributedWinners`: one which would retrieve the replicates, the other obtaining it from the parameters. Both version could be added to all `getNbReplicates...`. I guess they all could be potential race-condition issues.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Why not adding `ReplicatesList` retrieval in `runningTasks.stream()` filter for now?

mcornaton: We need the freshest value of `ReplicatesList` in the `while` loop, so it has to be retrieved when the task is handled, not before.

jeremyjams: Yeah, I understand, but this method needs at least couple updates if we don't want to move everything in DB query for this PR:
1. `isTeeTask && !worker.isTeeEnabled()` can be done with DB query
2. `hasWorkerAlreadyParticipated()` should be static
3. `isInContributionPhase` is useless if code is fast
4. `smsService.getEnclaveChallenge(chainTaskId, isTeeTask)` call can be moved outisde plus sms service should cache enclave challenges to avoid repeating calls on same tasks

mcornaton: I agree with you, there's still a lot to do on that method and the methods it calls. I'd say:
- every DB query that's needed to compute whether the worker can take the task or not should be done at the start of the `while` loop so that's fixed throughout the computation;
- consequence: every called method after the start of the `while` loop should be static.
Problem: using static methods is usually a bad practice. In particular, it makes things harder to unit test.

So, regarding your proposals:
1. Not sure what you mean. Do you mean the worker's state should be retrieved when needed - probably at the start of the `while` loop instead of before the loop? By the way, there's a TODO there: `access worker TEE status through workerService to avoid workerService.getWorker call`.
2. I don't think it'd be a good idea because of first paragraph. But it could be relieved from its call to the database.
3. Code is not always fast, remember @jbern0rd's work on fixing some `TASK_NOT_ACTIVE` issues. It usually depends on the load on the Scheduler. I don't think it'd be a good idea to remove that line.
4. Totally true. Let's use some memoization there!

To conclude: there's a lot to do to avoid race conditions there. We should probably also strive to propose some non-database-dependent methods in `ReplicateService` so that we would tempted to not use the database if not needed in future developments. See https://github.com/iExecBlockchainComputing/iexec-core/pull/490#discussion_r786713900 for an example.

src/main/java/com/iexec/core/sms/SmsService.java
```diff
@@ -20,25 +20,40 @@
...
+
+import static com.iexec.core.task.Task.LONGEST_TASK_TIMEOUT;
+
 
 @Slf4j
 @Service
 public class SmsService {
+    /**
+     * Memoize enclave challenges as they are computed by the SMS.
+     * They are constant over time, so we don't need to compute them everytime.
```
jeremyjams, line 39: :+1:

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: I see your point! Let's add this :)

src/main/java/com/iexec/core/task/TaskUpdateManager.java
mcornaton: Getting back to this: while I still agree with what I said in my last comment, having a `MetaTask` object could be great. It'd be computed at the start of `TaskUpdateManager#updateTask` and passed down to all `status12status2` methods in `TaskUpdateManager`. That'd help ensuring there's no desynchronization when we are for example in this block:
```
case RUNNING:
  running2ConsensusReached(task);
  running2RunningFailed(task);
  initializedOrRunning2ContributionTimeout(task);
  break;
```
This would go with the other optimizations I mentioned [there](https://github.com/iExecBlockchainComputing/iexec-core/pull/490#discussion_r786713900).

I'd like to work on this point once https://github.com/iExecBlockchainComputing/iexec-core/pull/484 is merged.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: Added in https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/50cb765c8117cb98b63fcbceddbbafc408e13d44.

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -110,7 +111,16 @@ public TaskService(TaskRepository taskRepository) {
     }
 
     public List<Task> getInitializedOrRunningTasks() {
-        return taskRepository.findByCurrentStatus(Arrays.asList(INITIALIZED, RUNNING),
+        return getInitializedOrRunningTasks(false);
+    }
+
+    public List<Task> getInitializedOrRunningTasks(boolean onlyStandardTasks) {
```
jeremyjams, line 117: `onlyStandardTasks` should fetch "any but not 0x0..01"
Else "any" is fetched

jeremyjams: To be clearer, with current implementation, a TEE-enabled worker will never be able to compute standard tasks (but it should).

mcornaton: Oh, yes, you're right. My bad!

jeremyjams: Thanks! Do you think `boolean onlyStandardTasks` -> `boolean shouldExcludeTeeTasks` is better?

src/main/java/com/iexec/core/task/TaskService.java
```diff
@@ -110,7 +111,27 @@ public TaskService(TaskRepository taskRepository) {
...
+     * then only standard tasks are retrieved.
+     * Otherwise, all tasks are retrieved.
+     *
+     * @param onlyStandardTasks Whether TEE tasks should be retrieved
+     *                          as well as standard tasks.
+     * @return A list of tasks which are {@link TaskStatus#INITIALIZED}
+     * or {@link TaskStatus#RUNNING}
+     */
+    public List<Task> getInitializedOrRunningTasks(boolean onlyStandardTasks) {
+        final String noTeeTag = onlyStandardTasks
```
jeremyjams, line 129:
```suggestion
        final String excludedTag = ...
```

src/main/java/com/iexec/core/task/TaskService.java
mcornaton: That may be better, yes. It represents the operation in a better way. Thanks for the idea!

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -162,14 +163,14 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
             final boolean taskNeedsMoreContributions = ConsensusHelper.doesTaskNeedMoreContributionsForConsensus(
                     chainTaskId,
-                    replicates,
+                    replicatesList.getReplicates(),
                     task.getTrust(),
                     task.getMaxExecutionTime());
 
             if (taskNeedsMoreContributions) {
+                boolean isTeeTask = task.isTeeTask();
                 String enclaveChallenge = smsService.getEnclaveChallenge(chainTaskId, isTeeTask);
```
jeremyjams, line 174: `smsService.getEnclaveChallenge(chainTaskId, isTeeTask)` can be:
1. Totally moved to `TaskUpdateManager.received2Initializing` and stored in `task.enclaveChallenge` for later use
2. (In order to honor your cache) Called a first time in `TaskUpdateManager.received2Initializing`, to relocate real HTTP call outside this method.

jeremyjams:
Here is a simplistic vision of what could be imagined in order to fetch fresher tasks from the database:
```
@Retryable(value = {OptimisticLockingFailureException.class}, maxAttempts = 5)
Optional<WorkerpoolAuthorization> getAuthOfAvailableReplicate(long workerLastBlock, String walletAddress) {
    List<String> alreadyScannedTasks = new ArrayList<>();

    //[...]

    Worker worker = workerService.getWorker(walletAddress).orElse(null);

    //[...]

    return findAvailableReplicateForWorker(worker, alreadyScannedTasks);
}

private Optional<WorkerpoolAuthorization> findAvailableReplicateForWorker(Worker worker, List<String> alreadyScannedTasks) {
    String chainTaskId = getNextInitializedOrRunningTaskId(!worker.isTeeEnabled(), alreadyScannedTasks);
    if (chainTaskId == null){
        // no work available
        return Optional.empty();
    }
    Task task = taskService.getTaskByChainTaskId(chainTaskId).orElse(null);

    //[...]

    boolean hasWorkerAlreadyParticipated = replicatesService.hasWorkerAlreadyParticipated(
            replicatesList, worker);
    if (hasWorkerAlreadyParticipated){
        // work cannot be scheduled on current worker
        alreadyScannedTasks.add(chainTaskId);
        return findAvailableReplicateForWorker(worker, alreadyScannedTasks);
    }
    //[...]

    // generate contribution authorization
    return Optional.of(signatureService.createAuthorization(
            worker, chainTaskId, enclaveChallenge));
}

private String getNextInitializedOrRunningTaskId(boolean shouldExcludeTeeTask, List<String> alreadyScannedTasks) {
    // return empty if there is no task to contribute
    List<String> runningTaskIds = taskService.getInitializedOrRunningTasks(shouldExcludeTeeTask)
            .stream()
            .map(Task::getChainTaskId)
            .collect(Collectors.toList());
    runningTaskIds.removeAll(alreadyScannedTasks);
    return runningTaskIds.get(0);
}
```

mcornaton:
I agree with you on the fact that `getAuthOfAvailableReplicate` should be refactored. However, I have a few comments on your proposition:
- in `getNextInitializedOrRunningTaskId`, you'll fetch and loop through all `INITIALIZED` or `RUNNING` task, which may represent a big load for the DB (retrieval) and for the Scheduler (loop). I think it'd be better to defer this filtering to the DB and retrieve tasks one after the other. e.g. in pseudo-SQL:
```
SELECT *
FROM Task
WHERE currentStatus IN (`INITIALIZED`, `RUNNING`) 
  AND chainTaskId NOT IN ($alreadyScanned)
LIMIT 1
```
- I'm also wondering if the `hasWorkerAlreadyParticipated` check could be done in the query or if it would require more time/more load. I honestly don't know much about MongoDB's "join" complexity. If it'd better to join in the query, it could be improved to something like:
```
...
  AND $walletAddress NOT IN (SELECT walletAddress FROM Replicate WHERE Replicate.chainTaskId = Task.chainTaskId)
...
```
- I really, really don't like the modification of an input parameter (`alreadyScannedTasks.add(chainTaskId)`). IMHO, it's error-prone ([see Pure Functions on Wikipedia](https://en.wikipedia.org/wiki/Pure_function)). So we'd have to find another solution to track already scanned tasks.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: Excellent!

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
jeremyjams: I suggest we swap from:
```
updateTaskStatusAndSave(task, INITIALIZED, null);
replicatesService.createEmptyReplicateList(task.getChainTaskId());
```
to:
```
replicatesService.createEmptyReplicateList(task.getChainTaskId());
updateTaskStatusAndSave(task, INITIALIZED, null);
```
and update current comment (`// If there's no replicatesList`) to express empty optional for ReplicateList cannot happen, check is only here to prevent `'Optional.get()' without 'isPresent()' check`. warning.

mcornaton: Sure, good idea. See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/802aada1c3f0ed55ebf320170f8200d59f48c63c.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -118,39 +126,33 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
-                continue;
-            }
-
             taskService.initializeTaskAccessForNewReplicateLock(chainTaskId);
             if (taskService.isTaskBeingAccessedForNewReplicate(chainTaskId)) {
                 continue;//skip task if being accessed
             }
             taskService.lockTaskAccessForNewReplicate(chainTaskId);
 
             boolean hasWorkerAlreadyParticipated = replicatesService.hasWorkerAlreadyParticipated(
```
jeremyjams, line 154: Getting very nice!

I feel we can merge in a single `if` following items:
* isTaskBeingAccessedForNewReplicate
* isInContributionPhase
* hasWorkerAlreadyParticipated
* taskNeedsMoreContributions
* enclaveChallenge.isEmpty()

mcornaton: I'm not sure it'd be great. If we do as you suggest, either we'll have to pre-compute everything so that it'll represent some load, either we'll have to have a long `if` clause. I don't really like any of these cases. As that could be a critical part of our workflow, it should probably be the fastest possible.
I think the best thing to do there is to extract all these checks into another method (`boolean canWorkerAcceptTask(Task task, String walletAddress)`?).

mcornaton: So, I've tried something in https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/2468dc1dde1b5dba83de7dbb546b5876aaf6e7f8. That's a first step toward [your suggestion about refactoring `getAuthOfAvailableReplicate`](https://github.com/iExecBlockchainComputing/iexec-core/pull/490#issuecomment-1017601472).
Let me know what you think about it!

jeremyjams: Nice!

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -126,68 +112,120 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
+        // Check if task is still in contribution phase with INITIALIZED or RUNNING status
+        Optional<Task> upToDateTask = taskService.getTaskByChainTaskId(chainTaskId);
+        final boolean taskNeedsMoreContributions = ConsensusHelper.doesTaskNeedMoreContributionsForConsensus(
+                chainTaskId,
+                replicatesList.getReplicates(),
+                task.getTrust(),
+                task.getMaxExecutionTime());
+
+        if (upToDateTask.isEmpty()
+                || !TaskStatus.isInContributionPhase(upToDateTask.get().getCurrentStatus())
```
jeremyjams, line 218: I don't think `isInContributionPhase` is still relevant since `isConsensusReached` already checks on-chain status

src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -44,6 +44,21 @@
     @Query("{ 'currentStatus': {$in: ?0} }")
     List<Task> findByCurrentStatus(List<TaskStatus> statuses, Sort sort);
 
+    @Query("{ 'currentStatus': {$in: ?0} , 'tag': {$ne: ?1} }")
+    List<Task> findByCurrentStatusInAndTagNot(List<TaskStatus> statuses, String excludedTag, Sort sort);
+
+    // Shortcut for `findFirstByCurrentStatusInAndTagNotAndChainTaskIdNotIn`
+    default Optional<Task> findFirstTask(List<TaskStatus> statuses, String excludedTag, List<String> excludedChainTaskIds, Sort sort) {
```
jeremyjams, line 51: `findOneLikelyAvailableTask`
?
Do we need default?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -126,68 +112,120 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
+     * {@link Optional#empty()} otherwise.
+     */
+    private Optional<WorkerpoolAuthorization> getAuthorizationForAnyTask(
+            String walletAddress,
+            boolean isTeeEnabled) {
+        final List<String> alreadyScannedTasks = new ArrayList<>();
+
+        Optional<WorkerpoolAuthorization> authorization = Optional.empty();
+        while (authorization.isEmpty()) {
+            final Optional<Task> oTask = taskService.getFirstInitializedOrRunningTask(!isTeeEnabled, alreadyScannedTasks);
```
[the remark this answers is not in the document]
mcornaton: I disagree with you, I think it really matters as we want to minimize randomness there. Explicitly mentioning we'll look for the first task seems accurate.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -126,68 +112,120 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
...
-        return Optional.empty();
+        if (!taskAccessForNewReplicateLock.lockIfPossible(chainTaskId)) {
+            return false;
+        }
+
+        // no need to go further if the consensus is already reached on-chain
+        // the task should be updated since the consensus is reached but it is still in RUNNING status
+        if (taskUpdateManager.isConsensusReached(replicatesList)) {
+            taskAccessForNewReplicateLock.unlock(chainTaskId);
+            taskUpdateManager.publishUpdateTaskRequest(chainTaskId);
```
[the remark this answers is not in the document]
mcornaton: You likely to be right :)

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: Absolutely right, thanks!

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -126,68 +112,120 @@ public ReplicateSupplyService(ReplicatesService replicatesService,
         }
         Worker worker = optional.get();
 
-        for (Task task : validTasks) {
-            String chainTaskId = task.getChainTaskId();
+        return getAuthorizationForAnyTask(
```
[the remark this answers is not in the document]
mcornaton: Thanks!

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: I still prefer to keep `first` in the method name as results are sorted before the first one is returned. BTW, I also prefer to stay generic as this does not simply find available tasks, it finds tasks according to criteria. Maybe something like `findFirstTaskGivenCritiria`?

We need `default` as this is a method whose body is defined in an interface.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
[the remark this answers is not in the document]
jeremyjams: Yes, I see your point, the problem I have is this method name seems to always return the task which has been initialized the first one (== always returns the same task ..!), but maybe it's just me..!

src/main/java/com/iexec/core/task/TaskRepository.java
jeremyjams: If you really want to add a alias on this method (`findFirstTaskGiven`) I suggest you attach a full javadoc

mcornaton: I really think it's better to have an alias, yes. The other method name is really, really long so providing an alias is mandatory IMHO. I'll add some JavaDoc to explain in-depth what's exactly this alias.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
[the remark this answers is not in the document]
mcornaton: I understand your concern and didn't think about that before. What do you think about `getPrioritizedInitializedOrRunningTask`?

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: Regarding the naming: what about using `prioritized` there, too? Something like `findPrioritizedTask`?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/36012c0fead204bc2994a77e71b7d8a98d4e2fac.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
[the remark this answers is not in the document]
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/36012c0fead204bc2994a77e71b7d8a98d4e2fac.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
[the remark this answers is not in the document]
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/36012c0fead204bc2994a77e71b7d8a98d4e2fac.

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/e02f7a6bca5fefefe56b1b901f9c5fdc0418d7bc.

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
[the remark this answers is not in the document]
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/e02f7a6bca5fefefe56b1b901f9c5fdc0418d7bc.

jeremyjams:
Great :)

src/main/java/com/iexec/core/tools/ContextualLock.java
```diff
@@ -0,0 +1,92 @@
...
+     * Gets a lock on a particular key if no lock currently exists on that key.
+     * <p>
+     * This method is synchronized on the key,
+     * so we're sure there shouldn't be any race condition.
+     *
+     * @param key Key to lock on.
+     * @return {@literal true} if the lock has been acquired,
+     * {@literal false} otherwise.
+     */
+    public boolean lockIfPossible(K key) {
```
jeremyjams, line 50: What about adding at least one UT with a 1st locking thread + a 2nd one trying to lock (but fail) without waiting. E.g I see a thread2 which loops by doing lockIfPossible which always fail but only succeeds right after thread1 unlocks.

mcornaton: Good idea. See https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/009e861c03e999070b21aa8443d1563ef2e49a4a.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
jbern0rd: Mocking static methods with mockito seems now possible.
See https://www.baeldung.com/mockito-mock-static-methods
The prerequisite is mockito 3.4.0 and spring framework fetches mockito 4.0.0

jbern0rd:
Is the `@Retryable` annotation still at the right place in `ReplicateSupplyService` ?

src/main/java/com/iexec/core/replicate/ReplicateSupplyService.java
```diff
@@ -28,11 +28,11 @@
 import com.iexec.core.chain.Web3jService;
 import com.iexec.core.contribution.ConsensusHelper;
 import com.iexec.core.detector.task.ContributionTimeoutTaskDetector;
```
jbern0rd, line 30: Unused import

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Good to know! So, let's make all these methods static with a non-static version when needed.

src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
```diff
@@ -78,16 +75,18 @@
     @Mock private TaskService taskService;
     @Mock private TaskUpdateRequestManager taskUpdateRequestManager;
     @Mock private WorkerService workerService;
-    @Mock private SmsService smsService;
     @Mock private Web3jService web3jService;
     @Mock private ContributionTimeoutTaskDetector contributionTimeoutTaskDetector;
```
jbern0rd, line 79: This detector is no more used in `ReplicateSupplyService`.
Could we clean this test class ?

mcornaton: Sure, done in https://github.com/iExecBlockchainComputing/iexec-core/pull/490/commits/b3f5a3128f808b0aa595a83ad09594810d96e7de.

jbern0rd (approved):

src/main/java/com/iexec/core/replicate/ReplicatesHelper.java
```diff
@@ -0,0 +1,131 @@
...
+     * that have the right contribution hash.
+     * <p>
+     * Note this method won't retrieve the replicates list, so it could be static.
+     * For test purposes - i.e. mocking -, it is not.
+     *
+     * @param replicates       List of replicates on which the calculation should be base.
+     * @param contributionHash Valid hash of the final result.
+     * @return Number of winners who had contributed with the right hash.
+     */
+    public static int getNbValidContributedWinners(List<Replicate> replicates, String contributionHash) {
```
zguesmi, line 43: Why not putting these methods (methods with `List<Replicate>` as argument) in `ReplicateList`?

src/main/java/com/iexec/core/replicate/ReplicatesHelper.java
```diff
@@ -0,0 +1,131 @@
...
+            for (ReplicateStatus status : listStatus) {
+                if (listReplicateStatus.contains(status)) {
+                    addressReplicates.add(replicate.getWalletAddress());
+                }
+            }
+        }
+        return addressReplicates.size();
+    }
+
+    static boolean isStatusBeforeWorkerLostEqualsTo(Replicate replicate, ReplicateStatus status) {
```
zguesmi, line 94: Wouldn't be better to put this method in `Replicate`?

src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -44,6 +44,42 @@
     @Query("{ 'currentStatus': {$in: ?0} }")
     List<Task> findByCurrentStatus(List<TaskStatus> statuses, Sort sort);
 
+    @Query("{ 'currentStatus': {$in: ?0} , 'tag': {$ne: ?1} }")
```
zguesmi, line 47: I think there is no need for this as spring will parse the current name of the method and reconstruct the request.

src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -44,6 +44,42 @@
...
+     * </ul>
+     *
+     * @param statuses             The task status should be one of this list.
+     * @param excludedTag          The task tag should not be this tag
+     *                             - use {@literal null} if no tag should be excluded.
+     * @param excludedChainTaskIds The chain task ID should not be one of this list.
+     * @param sort                 How to prioritize tasks.
+     * @return The first task matching with the criteria, according to the {@code sort} parameter.
+     */
+    default Optional<Task> findPrioritizedTask(List<TaskStatus> statuses,
```
zguesmi, line 69: Why not in `TaskService`? For me the repository should contain the least amount of logic and only define the interface with repositories.

zguesmi (approved):
Despite the minor comments, I find the PR excellent, thanks @mcornaton. My comments are just recommendations, they shouldn't prevent you from merging the PR. If we all agree on the suggested enhancements w could do them in another one.

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: The only point of `findPrioritizedTask` is to avoid using `findFirstByCurrentStatusInAndTagNotAndChainTaskIdNotIn` outside of this repo. This method's name is way too long so I supposed it would be better to add an alias directly into the repo.

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: You're right. Let's remove it!

src/main/java/com/iexec/core/task/TaskRepository.java
zguesmi: If we move `findPrioritizedTask` to the service, the long-named method will only be called once. Other services should call `TaskService`.

src/main/java/com/iexec/core/replicate/ReplicatesHelper.java
mcornaton: Sure!

src/main/java/com/iexec/core/replicate/ReplicatesHelper.java
mcornaton: Sure!

src/main/java/com/iexec/core/task/TaskRepository.java
mcornaton: Alright, see https://github.com/iExecBlockchainComputing/iexec-core/pull/496 :)

PR #491, opened 2022-01-11
Update technical debt

mcornaton:


src/test/java/com/iexec/core/chain/DealWatcherServiceTests.java
```diff
@@ -56,7 +56,7 @@
 
     @BeforeEach
     public void init() {
```
jeremyjams, line 58: If you like you can `Remove this 'public' modifier.`

jeremyjams (commented):
Great!

src/test/java/com/iexec/core/chain/DealWatcherServiceTests.java
mcornaton: Sure, let's remove all useless `public` modifiers of tests methods/classes!

jeremyjams (approved):

src/main/java/com/iexec/core/workflow/Workflow.java
```diff
@@ -86,12 +85,11 @@ void removeTransition(T from, T to) {
...
     public boolean isValidTransition(T from, T to){
         return possibleTransitions.containsKey(from)
-            ? possibleTransitions.get(from).contains(to)
-            : false;
+                && possibleTransitions.get(from).contains(to);
     }
 
     List<T> toList(T... statuses) {
-        return new ArrayList<T>(Arrays.asList(statuses));
+        return new ArrayList<>(Arrays.asList(statuses));
```
jbern0rd, line 92:
```suggestion
        return List.of(statuses);
```
Is this method really necessary ?

jbern0rd (approved):

src/main/java/com/iexec/core/workflow/Workflow.java
mcornaton: I don't think so. I'll add a `@Deprecated` annotation. Thanks for pointing it out.

mcornaton: Well, I checked the code and we finally need it. The list should be [mutable](https://github.com/iExecBlockchainComputing/iexec-core/blob/91bf3e19ed8f8d0f4248ac7084b20459e390b68f/src/main/java/com/iexec/core/workflow/Workflow.java#L44), so that's probably better to keep this method as is.

jbern0rd: I missed this, thanks for the counter check

Issue #482, opened 2021-12-23
Replicate status concurrent update


mcornaton:
When a detector detects a missing update and the worker notifies the scheduler at the same time for the same update, both updates will be executed, leading to some incoherent state. Such as:
```
    Expecting:
     <[CREATED,
        STARTING,
        STARTED,
        APP_DOWNLOADING,
        APP_DOWNLOADED,
        DATA_DOWNLOADING,
        DATA_DOWNLOADED,
        COMPUTING,
        COMPUTED,
        CONTRIBUTING,
        CONTRIBUTED,
        CONTRIBUTED,    <--- CONTRIBUTED x2 ??
        REVEALING,
        REVEALED,
        COMPLETING,
        COMPLETED]>
    to be equal to:
     <[CREATED,
        STARTING,
        STARTED,
        APP_DOWNLOADING,
        APP_DOWNLOADED,
        DATA_DOWNLOADING,
        DATA_DOWNLOADED,
        COMPUTING,
        COMPUTED,
        CONTRIBUTING,
        CONTRIBUTED,
        REVEALING,
        REVEALED,
        COMPLETING,
        COMPLETED]>
```

This would probably need some synchronization on `ReplicatesService#updateReplicateStatus` based on worker's wallet address.

PR #492, opened 2022-01-13
Fix race condition on replicate update

mcornaton:
Depends on https://github.com/iExecBlockchainComputing/iexec-common/pull/295.

### DIFF
--- a/src/main/java/com/iexec/core/replicate/ReplicatesService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicatesService.java
@@ -414,6 +414,19 @@ public class ReplicatesService {
      *   3) if worker did succeed onChain when CONTRIBUTED/REVEALED.
      *   4) if worker did upload when RESULT_UPLOADING.
      */
+
+    /**
+     * This method updates a replicate while caring about thread safety.
+     * A single replicate can then NOT be updated twice at the same time.
+     * This method should be preferred to
+     * {@link ReplicatesService#updateReplicateStatusWithoutThreadSafety(String, String, ReplicateStatusUpdate, UpdateReplicateStatusArgs)}!
+     *
+     * @param chainTaskId Chain task id of the task whose replicate should be updated.
+     * @param walletAddress Wallet address of the worker whose replicate should be updated.
+     * @param statusUpdate Info about the status update - new status, date of update, ...
+     * @param updateReplicateStatusArgs Optional args used to update the status.
+     * @return An optional next action for the worker.
+     */
     @Retryable(value = {OptimisticLockingFailureException.class}, maxAttempts = 100)
     public Optional<TaskNotificationType> updateReplicateStatus(String chainTaskId,
                                                                 String walletAddress,
@@ -428,7 +441,23 @@ public class ReplicatesService {
         );
     }
 
-    private Optional<TaskNotificationType> updateReplicateStatusWithoutThreadSafety(String chainTaskId, String walletAddress, ReplicateStatusUpdate statusUpdate, UpdateReplicateStatusArgs updateReplicateStatusArgs) {
+    /**
+     * This method updates a replicate but does not care about thread safety.
+     * A single replicate can then be updated twice at the same time
+     * and completely break a task.
+     * This method has to be used with a synchronization mechanism, e.g.
+     * {@link ReplicatesService#updateReplicateStatus(String, String, ReplicateStatus, ReplicateStatusDetails)}
+     *
+     * @param chainTaskId Chain task id of the task whose replicate should be updated.
+     * @param walletAddress Wallet address of the worker whose replicate should be updated.
+     * @param statusUpdate Info about the status update - new status, date of update, ...
+     * @param updateReplicateStatusArgs Optional args used to update the status.
+     * @return An optional next action for the worker.
+     */
+    private Optional<TaskNotificationType> updateReplicateStatusWithoutThreadSafety(String chainTaskId,
+                                                                                    String walletAddress,
+                                                                                    ReplicateStatusUpdate statusUpdate,
+                                                                                    UpdateReplicateStatusArgs updateReplicateStatusArgs) {
         log.info("Replicate update request [status:{}, chainTaskId:{}, walletAddress:{}, details:{}]",
                 statusUpdate.getStatus(), chainTaskId, walletAddress, statusUpdate.getDetailsWithoutStdout());
 
--- a/src/main/java/com/iexec/core/replicate/ReplicatesService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicatesService.java
@@ -23,7 +23,7 @@ import com.iexec.common.replicate.ReplicateStatusCause;
 import com.iexec.common.replicate.ReplicateStatusDetails;
 import com.iexec.common.replicate.ReplicateStatusUpdate;
 import com.iexec.common.task.TaskDescription;
-import com.iexec.common.utils.TargetedLock;
+import com.iexec.common.utils.ContextualLockRunner;
 import com.iexec.core.chain.IexecHubService;
 import com.iexec.core.chain.Web3jService;
 import com.iexec.core.result.ResultService;
@@ -55,8 +55,8 @@ public class ReplicatesService {
     private ResultService resultService;
     private StdoutService stdoutService;
 
-    private final TargetedLock<String> replicatesUpdateLock =
-            new TargetedLock<>(10, TimeUnit.MINUTES);
+    private final ContextualLockRunner<String> replicatesUpdateLock =
+            new ContextualLockRunner<>(10, TimeUnit.MINUTES);
 
     public ReplicatesService(ReplicatesRepository replicatesRepository,
                              IexecHubService iexecHubService,
--- a/src/main/java/com/iexec/core/replicate/ReplicatesService.java
+++ b/src/main/java/com/iexec/core/replicate/ReplicatesService.java
@@ -380,6 +380,15 @@ public class ReplicatesService {
     }
 
     @Recover
+    public void updateReplicateStatus(OptimisticLockingFailureException exception,
+                                      String chainTaskId,
+                                      String walletAddress,
+                                      ReplicateStatus newStatus,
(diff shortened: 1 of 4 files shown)


src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -423,6 +419,15 @@ public void updateReplicateStatus(OptimisticLockingFailureException exception,
                                                                 String walletAddress,
                                                                 ReplicateStatusUpdate statusUpdate,
                                                                 UpdateReplicateStatusArgs updateReplicateStatusArgs) {
+        // Synchronization is mandatory there to avoid race conditions.
+        // Lock key should be unique, e.g. `chainTaskId + walletAddress`.
+        return replicatesUpdateLockRunner.getWithLock(
+                chainTaskId + walletAddress,
```
zguesmi, line 425:
```suggestion
        String lockKey = chainTaskId + walletAddress;
        return replicatesUpdateLockRunner.getWithLock(
                key,
```

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -423,6 +419,15 @@ public void updateReplicateStatus(OptimisticLockingFailureException exception,
...
+                () -> updateReplicateStatusWithoutSync(chainTaskId, walletAddress, statusUpdate, updateReplicateStatusArgs)
+        );
+    }
+
+    private Optional<TaskNotificationType> updateReplicateStatusWithoutSync(String chainTaskId, String walletAddress, ReplicateStatusUpdate statusUpdate, UpdateReplicateStatusArgs updateReplicateStatusArgs) {
```
zguesmi, line 430: What do you mean by "without sync" here? The method itself is not synchronized?

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Done in https://github.com/iExecBlockchainComputing/iexec-core/pull/492/commits/6361f1cb2658065355f4fac4edd2f7ce5452055a.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: That's it. I wanted to add some clear difference with `updateReplicateStatus`. That may not be the right wording, I'm not sure at all.

jeremyjams: Something like `updateReplicateStatusNow`? (I'm sure you have something better !)

zguesmi: We can keep the name of this method as is, and change the calling method (the retryable one) to `updateReplicateStatusAsync`.

mcornaton: @zguesmi > The calling method isn't async, so it doesn't work :/
The idea is to tell in the method name that it could be subject to race conditions. But I don't find any short and clear ways to do so.

zguesmi: Technically it is not async indeed, but practically it is running the update asynchronously.

mcornaton: I don't get it. There's no asynchronously update there. Everything is synchronous in `updateReplicateStatus`, `updateReplicateStatusWithoutSync` and `ContextualLockRunner`.

jeremyjams: @mcornaton I guess you don't want to rename L418 to
`public Optional<TaskNotificationType> updateReplicateStatusThreadSafely`
to keep a short name from the caller part?

mcornaton: Yes, that's my concern. But I still don't have any better solution :thinking:

zguesmi: The day we signed on to work with java, there was a hidden statement, "iAgreeToUsingVeryLongLongNamesInJavaDespiteThemBeingVeryLong" :smile:

zguesmi: Joking apart, I don't find the name `updateReplicateStatusThreadSafely` to be that long.
The other option would be to rename `updateReplicateStatusWithoutSync` to `updateReplicateStatusWithoutThreadSafety` (for example) with a comment explaining why and which method to use in which case.

src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
```diff
@@ -659,6 +660,32 @@ public void shouldNotUpdateReplicateStatusSinceAlreadyReported() {
...
+        when(iexecHubService.repeatIsRevealedTrue(anyString(), anyString())).thenReturn(true);
+        when(replicatesRepository.save(replicatesList)).thenReturn(replicatesList);
+
+        ReplicateStatusUpdate statusUpdate = ReplicateStatusUpdate.builder()
+                .modifier(WORKER)
+                .status(REVEALED)
+                .build();
+
+        // Without any synchronization mechanism,
+        // this would update 10 times to `REVEALED`.
```
zguesmi, line 680: I think it would be interesting to run the same thing (10 updates) on the non-thread-safe method and assert that the wrong updates happened in because of concurrent threads (update to REVEALED more than once). That would highlight the efficiency of the new mechanism.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: I prefer the second option as it should be used a single time ; it won't add lots of characters overall.

src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
mcornaton: Good idea, thanks! See https://github.com/iExecBlockchainComputing/iexec-core/pull/492/commits/949f44a332fbf89f6ef13998726f1d302096c581.

zguesmi (approved):

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -418,11 +414,50 @@ public void updateReplicateStatus(OptimisticLockingFailureException exception,
...
+     * {@link ReplicatesService#updateReplicateStatusWithoutThreadSafety(String, String, ReplicateStatusUpdate, UpdateReplicateStatusArgs)}!
+     *
+     * @param chainTaskId Chain task id of the task whose replicate should be updated.
+     * @param walletAddress Wallet address of the worker whose replicate should be updated.
+     * @param statusUpdate Info about the status update - new status, date of update, ...
+     * @param updateReplicateStatusArgs Optional args used to update the status.
+     * @return An optional next action for the worker.
+     */
     @Retryable(value = {OptimisticLockingFailureException.class}, maxAttempts = 100)
     public Optional<TaskNotificationType> updateReplicateStatus(String chainTaskId,
```
jbern0rd, line 431: Just to understand: calls on locks are non-blocking and we try a bunch of times until the lock is obtained ? (A try every second for 100 seconds)
When the deadline is reached, we go to the @Recover annotated method.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -464,7 +499,7 @@ public void updateReplicateStatus(OptimisticLockingFailureException exception,
                         "nextAction:{}, chainTaskId:{}, walletAddress:{}]",
                 replicate.getCurrentStatus(), newStatusCause, nextAction, chainTaskId, walletAddress);
 
-        return Optional.ofNullable(nextAction); // should we return a default action when null?
+        return Optional.ofNullable(nextAction);
     }
 
     @Recover
```
jbern0rd, line 505: Could the `chainTaskId`, `wallet address` and `status update` be written to the log ?
Could we have `log.error("message", exception)` in the log method instead of `printStackTrace()` ?

jbern0rd: I am speaking of `updateReplicateStatus(OptimisticLockingFailureException exception, ...)` and `logUpdateReplicateStatusRecover(OptimisticLockingFailureException exception)` methods

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: Calls on locks are blocking.
The `@Retryable` annotation helps with database issues - the database mechanism also uses some locking feature. So we try 100 times if the database fails to save the new replicate status.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
mcornaton: As far as I remember, those values are displayed in the error message. I'm not sure it would add any value to duplicate them (?).
However, second point is right. See https://github.com/iExecBlockchainComputing/iexec-core/pull/492/commits/388a3d19ef57dfd745f7b5ebe6d51b5d52f26377.

mcornaton (commented):
https://github.com/iExecBlockchainComputing/iexec-core/pull/492/commits/042d11aa90ae085c61cb24fc830b1e67cbe46cc1: Note that `@Retryable` should ideally be put on `updateReplicateStatusWithoutThreadSafety` but that won't work with Spring's Retryable.

src/main/java/com/iexec/core/replicate/ReplicatesService.java
```diff
@@ -403,8 +408,7 @@ public void updateReplicateStatus(String chainTaskId,
     public void updateReplicateStatus(OptimisticLockingFailureException exception,
```
jbern0rd, line 408: The `@Recover` (`void`) and `@Retryable` (`Optional`) methods don't have the same return type.
It seems weird.

jbern0rd (approved):
Great work, thanks for the fix !
Let's see this running

mcornaton:
Fix #482.

PR #493, opened 2022-01-21
Catch `ConcurrentModificationException` in tests

mcornaton:
A case was missing on thread non-safety test.

### DIFF
--- a/src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
+++ b/src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
@@ -702,9 +702,17 @@ class ReplicateServiceTests {
 
         // Without any synchronization mechanism,
         // this would update between 1 and 10 times to `REVEALED`.
-        IntStream.range(0, 10)
-                .parallel()
-                .forEach(i -> replicatesService.updateReplicateStatusWithoutThreadSafety(CHAIN_TASK_ID, WALLET_WORKER_1, statusUpdate, UPDATE_ARGS));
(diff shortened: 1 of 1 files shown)


jeremyjams (approved):

src/test/java/com/iexec/core/replicate/ReplicateServiceTests.java
```diff
@@ -702,9 +702,17 @@ void shouldEncounterRaceConditionOnReplicateUpdateWithoutThreadSafety() {
...
+        // Or this could throw a `ConcurrentModificationException`
+        // on `Replicate#containsStatus` call.
+        try {
+            IntStream.range(0, 10)
+                    .parallel()
+                    .forEach(i -> replicatesService.updateReplicateStatusWithoutThreadSafety(CHAIN_TASK_ID, WALLET_WORKER_1, statusUpdate, UPDATE_ARGS));
+        } catch (ConcurrentModificationException e) {
+            System.out.println("Concurrent modification detected," +
+                    " thread safety is effectively not met.");
+            return;
```
jbern0rd, line 714: Shouldn't the test fail if thread safety is not met ?
```suggestion
            Assertion.fail("Concurrent modification detected," +
                    " thread safety is effectively not met."));
```

jbern0rd: Forget it, if we have a ConcurrentModificationException, we have a race condition.
This is the purpose of the test

jbern0rd (approved):

PR #496, opened 2022-02-17
Enhance #PR 490

mcornaton:
Add enhancements to `ReplicateSupplyService#getAuthOfAvailableReplicate` refactor based on @zguesmi's comments.


jbern0rd (approved):
I find it much cleaner this way.
Thanks !

src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
```diff
@@ -233,24 +234,22 @@ void shouldNotGetAnyReplicateSinceWorkerAlreadyParticipated() {
...
         when(taskService.getPrioritizedInitializedOrRunningTask(true, Collections.emptyList()))
                 .thenReturn(Optional.of(runningTask));
         when(workerService.getWorker(WALLET_WORKER_1)).thenReturn(Optional.of(existingWorker));
         when(replicatesService.getReplicatesList(CHAIN_TASK_ID)).thenReturn(Optional.of(replicatesList));
 
-        try (MockedStatic<ReplicatesHelper> replicatesHelper = Mockito.mockStatic(ReplicatesHelper.class)) {
-            replicatesHelper.when(() -> ReplicatesHelper.hasWorkerAlreadyParticipated(replicatesList, WALLET_WORKER_1)).thenReturn(true);
-            Optional<WorkerpoolAuthorization> oAuthorization =
-                    replicateSupplyService.getAuthOfAvailableReplicate(workerLastBlock, WALLET_WORKER_1);
+        when(replicatesList.hasWorkerAlreadyParticipated(WALLET_WORKER_1)).thenReturn(true);
```
zguesmi, line 248: Are you sure `when(...).then(...)` works on `Spy` objects? I think we should use `do(...).when(...)` instead.

src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -44,11 +44,7 @@
     @Query("{ 'currentStatus': {$in: ?0} }")
     List<Task> findByCurrentStatus(List<TaskStatus> statuses, Sort sort);
 
-    @Query("{ 'currentStatus': {$in: ?0} , 'tag': {$ne: ?1} }")
-    List<Task> findByCurrentStatusInAndTagNot(List<TaskStatus> statuses, String excludedTag, Sort sort);
```
zguesmi, line 48: :+1:

src/test/java/com/iexec/core/replicate/ReplicateSupplyServiceTests.java
zguesmi: Good, fair enough. I'm pretty sure I had a different behavior before but maybe it wasn't the same thing or new versions changed it. Thanks for verifying that.

zguesmi (approved):

PR #499, opened 2022-03-23
Feature/use blockchain adapter api library

jbern0rd:


src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
```diff
@@ -128,44 +124,33 @@ public BlockchainAdapterService(BlockchainAdapterClient blockchainAdapterClient)
...
-            return Optional.empty();
-        }
-        ResponseEntity<CommandStatus> commandStatusEntity;
-        try {
-            commandStatusEntity = getCommandStatusFunction.apply(chainTaskId);
-            if (!commandStatusEntity.getStatusCode().is2xxSuccessful()
-                    || commandStatusEntity.getBody() == null) {
+            long period, int maxAttempts) {
+        int attempt = 0;
+        while(attempt < maxAttempts) {
```
mcornaton, line 135: Why not a `for` loop?

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
```diff
@@ -53,15 +53,13 @@ public BlockchainAdapterService(BlockchainAdapterClient blockchainAdapterClient)
...
         try {
-            ResponseEntity<String> initializeResponseEntity =
-                    blockchainAdapterClient.requestInitializeTask(chainDealId, taskIndex);
-            if (initializeResponseEntity.getStatusCode().is2xxSuccessful()
-                    && !StringUtils.isEmpty(initializeResponseEntity.getBody())) {
-                String chainTaskId = initializeResponseEntity.getBody();
-                log.info("Requested initialize [chainTaskId:{}, chainDealId:{}, " +
-                        "taskIndex:{}]", chainTaskId, chainDealId, taskIndex);
-                return Optional.of(chainTaskId);
+            String chainTaskId = blockchainAdapterClient.requestInitializeTask(chainDealId, taskIndex);
```
jeremyjams, line 56: Minor
```
        try {
            String chainTaskId [...]
            if (!StringUtils.isEmpty(chainTaskId)) {
                        [...]
                return Optional.of(chainTaskId);
            }
        } catch (Throwable e) {
            [...]
        }
        return Optional.empty();
```
Like `requestFinalize()`?

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
jeremyjams: Future note if it helps, a wrapped FailSafe example: https://github.com/iExecBlockchainComputing/iexec-common/blob/b9ea05659079397c65542eb54338697a7b5cb968/src/main/java/com/iexec/common/chain/IexecHubAbstractService.java#L878-L878

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
jbern0rd: fixed in 325115f

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
jbern0rd: while loop removed in 45fb7a9

mcornaton (approved):
LGTM! :smile:

src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java
```diff
@@ -53,16 +53,13 @@ public BlockchainAdapterService(BlockchainAdapterClient blockchainAdapterClient)
...
-                    && !StringUtils.isEmpty(initializeResponseEntity.getBody())) {
-                String chainTaskId = initializeResponseEntity.getBody();
+            String chainTaskId = blockchainAdapterClient.requestInitializeTask(chainDealId, taskIndex);
+            if (!StringUtils.isEmpty(chainTaskId)) {
                 log.info("Requested initialize [chainTaskId:{}, chainDealId:{}, " +
                         "taskIndex:{}]", chainTaskId, chainDealId, taskIndex);
                 return Optional.of(chainTaskId);
             }
-        } catch (Throwable e) {
+        } catch (Exception e) {
```
jeremyjams, line 62: `FeignException`? -> https://github.com/iExecBlockchainComputing/iexec-core/blob/325115f2281e587ab0564eced4ca4e36ae030a95/src/main/java/com/iexec/core/chain/adapter/BlockchainAdapterService.java#L165

jbern0rd: Are we sure that `FeignException` will be the only type of `RuntimeException` thrown ?
Wouldn't it be better to use `RutimeException` as an overall way to cacth most of happening exceptions during runtime ?

jeremyjams: 1. I thought so, but did not find the information in the doc, we can use the generic one if you prefer yes

PR #501, opened 2022-03-31
Feature/use result proxy library

jbern0rd:


gradle.properties
```diff
@@ -1,6 +1,7 @@
 version=7.0.1
 iexecCommonVersion=5.9.1-NEXT-SNAPSHOT
 iexecBlockchainAdapterVersion=7.0.0-NEXT-SNAPSHOT
+iexecResultVersion=7.0.2-NEXT-SNAPSHOT
```
jeremyjams, line 4: We might switch to a dedicated library version, which should be pretty stable over time. My point is: A new internal result-proxy patch should not make the previous API library version obsolete

zguesmi: Wouldn't the service (result-proxy) and its library progress together? Are you suggesting we dissociate them?

jbern0rd: I have several remarks on this:
* We have never used and published library this way, it's hard to understand the impacts
* If we switch to a dedicated library version, it breaks the idea to manage library and spring boot app in same repository (tags management will become too hard)

In the current topology, if a patch is done on a spring boot app, a new tag/version will be created and will create a new spring boot app as well as a new library. If the library has not been modified, the new library version will have the same functionality level as the previous one. We can stick to the previous library version or propagate it elsewhere

If the library has been modified, we will chose to deploy or not the new spring boot application and will need to use the library in other applications and potentially create new versions for those one.

If we want to split library management for spring boot application management, we must create dedicated repositories for  each libraries and adapt our workflow.

jeremyjams: I'm not saying to do it now, but we need to think about coupling we are introducing
Something like
iexecResultVersion=x.y.z
iexecResultApiVersion=${iexecResultVersion.major}.${iexecResultVersion.minor}

src/main/java/com/iexec/core/result/ResultService.java
```diff
@@ -55,18 +54,17 @@ public boolean isResultUploaded(String chainTaskId) {
             return false;
         }
 
-        return resultRepoClient.isResultUploaded(resultProxyToken, chainTaskId)
-                .getStatusCode()
-                .is2xxSuccessful();
+        resultProxyClient.isResultUploaded(resultProxyToken, chainTaskId);
```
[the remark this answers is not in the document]
jbern0rd: It is not about `isResultUploaded` being `false`, it is about its http status being 2xx or not.

On iexec-result-proxy side, the important part is here:
https://github.com/iExecBlockchainComputing/iexec-result-proxy/blob/develop/src/main/java/com/iexec/resultproxy/proxy/ProxyController.java#L117
We can only have a 2xx http status if result has been uploaded (http 204 no content).
It will be 404 not found or 401 unauthorized otherwise.

With the new feign clients, non-2xx http status will throw `FeignException` which will be managed with `spring-retry` annotations in iexec-core (`@Retryable` and `@Recover`).

Here, with spring-retry, when everything is ok, no exception happens and we can return true.
When errors happen several times in a row, we go to the recovery method and return a global false stating that upload failed several times and cannot be performed.
In both cases, we ignore the empty body and only return a boolean.

Note : No body is provided in iexec-result-proxy `ProxyController#isResultUploaded`.
The method signatures may be misleading.

src/main/java/com/iexec/core/result/ResultService.java
```diff
@@ -95,19 +93,22 @@ private String getResultProxyToken() {
     }
 
     private Optional<Eip712Challenge> getChallenge() {
-        ResponseEntity<Eip712Challenge> challengeResponse = resultRepoClient.getChallenge(chainConfig.getChainId());
-
-        if (challengeResponse != null && challengeResponse.getStatusCode().is2xxSuccessful()) {
-            return Optional.of(challengeResponse.getBody());
+        try {
```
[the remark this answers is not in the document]
jbern0rd: I see a problem here as the return outside of the catch block will never be reached by design.
Note : I just realized I missed something similar in a recent PR. I will try to find it a fix it soon.

We may try
```
try {
    Eip712Challenge challenge = getChallenge();
    // will produce nullpointerexception if challenge is null and go to the catch block
    return Optional.of(challenge);
} catch (FeignException e) {
    log.error("Failed to get challenge", e);
    return Optional.empty();
}
```

src/main/java/com/iexec/core/result/ResultService.java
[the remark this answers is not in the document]
jeremyjams: I just noticed `public ResponseEntity<**String**> isResultUploaded(..)` yes, this isn't meaningful to me yes

zguesmi (approved):

src/main/java/com/iexec/core/result/ResultService.java
```diff
@@ -95,19 +93,20 @@ private String getResultProxyToken() {
...
     }
 
     private String login(String token) {
-        ResponseEntity<String> loginResponse = resultRepoClient.login(chainConfig.getChainId(), token);
-        if (loginResponse != null && loginResponse.getStatusCode().is2xxSuccessful()) {
-            return loginResponse.getBody();
+        try {
+            return resultProxyClient.login(chainConfig.getChainId(), token);
+        } catch (RuntimeException e) {
+            return "";
```
jeremyjams, line 109: log?

jeremyjams (approved):

src/main/java/com/iexec/core/result/ResultService.java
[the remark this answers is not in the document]
jbern0rd: Note to myself, my comment above is false.
The suggested algorithm is valid.

mcornaton (approved):

PR #502, opened 2022-04-13
Bugfix/improve serialization deserialization

jbern0rd:


src/main/java/com/iexec/core/replicate/Replicate.java
```diff
@@ -36,6 +37,7 @@
 
 @Data
 @NoArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
```
jeremyjams, line 40: To better understand, in which cases (examples) `Replicate` is rebuilt with unwanted fields?

jbern0rd: IT is required for methods prefixed with `is` or `get` without and associated field.
To better understand, you can see the relation between commits and Jenkins jobs failures.
You will see the serialization errors, on which json fields they appeared and if those json fields were pojo fields or methods.

zguesmi: Does that mean, properties such as `isBusyComputing()` will not be present in the JSON anymore?

jbern0rd: `isBusyComputing` result will still be present, under the `busyComputing` key.
The `JsonIgnoreProperties` will kick during the deserialization, not the serialization.
As there is a isBusyComputing, which is assimilated to a getter, the framework looks up for a `busyComputing` field which does not exist in our model.

The added property allows to serialize (obtain in json) the return value of methods such as isBusyComputing, isRecoverable. When deserializing a json to a replicate, those data will be ignore. The framework will deserialize the ReplicateStatusUpdate list which contain all data required to compute again the aforementioned methods.

I will add a check on serialized json string next week to prove this.

zguesmi: Very clear, thanks!

jbern0rd: Test completed in d276081
You will see busyComputing is present in serialized JSON string

jeremyjams (approved):

zguesmi (approved):

PR #504, opened 2022-04-26
Task feedback

mcornaton:


jbern0rd (commented):
Is there a reason to introduce the new wording only for tasks and not for replicates ?

https://github.com/iExecBlockchainComputing/iexec-core/blob/develop/src/main/java/com/iexec/core/task/TaskController.java#L110

mcornaton:
Thanks for pointing this endpoint out. It's updated :)

@jeremyjams: Sure, I almost forgot about it. Let's work on this!

jbern0rd:
> Thanks for pointing this endpoint out. It's updated :)
> 
> @jeremyjams: Sure, I almost forgot about it. Let's work on this!

Quick feedback on this
docker logs command docker-java allows to retrieve both stdout and stderr
https://github.com/docker-java/docker-java/blob/master/docker-java-api/src/main/java/com/github/dockerjava/api/command/LogContainerCmd.java#L19

At the same time, we have
https://github.com/iExecBlockchainComputing/iexec-common/blob/develop/src/main/java/com/iexec/common/docker/DockerLogs.java

https://github.com/iExecBlockchainComputing/iexec-common/blob/develop/src/main/java/com/iexec/common/docker/client/DockerClientInstance.java#L735

We need to check worker correctly retrieves stdout and stderr for docker logs.

jbern0rd (approved):

src/main/java/com/iexec/core/task/TaskController.java
```diff
@@ -100,14 +100,20 @@ ReplicateModel buildReplicateModel(Replicate replicate) {
         return replicateModel;
     }
 
-    @GetMapping("/tasks/{chainTaskId}/stdout")
+    @GetMapping(path = {
+            "/tasks/{chainTaskId}/stdout",  // @Deprecated
+            "/tasks/{chainTaskId}/logs"
+    })
     public ResponseEntity<TaskStdout> getTaskStdout(@PathVariable("chainTaskId") String chainTaskId) {
```
jeremyjams, line 107:
```
TaskLogs
getTaskLogs
ReplicateLogs
...
*Stdout -> *Logs
```
?

mcornaton: We have some properties containing `stdout`:
```
stdout:
  purge-rate-in-days: ${IEXEC_STDOUT_PURGE_RATE_IN_DAYS:1}
  availability-period-in-days: ${IEXEC_STDOUT_AVAILABILITY_PERIOD_IN_DAYS:3}
```
I'm not sure we should update them as it could break external scripts/deployments.

mcornaton: BTW, if we want to later add `stderr`, it may be clearer to keep `stdout` there :thinking:

jeremyjams: Considering it is not likely clients use yml values directly but use environment variables, we might try that kind of things (?): 
```
logs:
  purge-rate-in-days: ${IEXEC_STDOUT_PURGE_RATE_IN_DAYS:${IEXEC_LOGS_PURGE_RATE_IN_DAYS:1}}
```
About your comment on stderr, we should target
```
logs {
  stdout
  stderr
}
```
When talking about purge rate, etc, log policy will be the same for both stdout and stderr (since we'll play with a single `logs` object).

What do you think?

mcornaton: Please see [Common commit](https://github.com/iExecBlockchainComputing/iexec-common/pull/326/commits/c171f751d38b879928839167df82adad4cbab3a9) and [Core commit](https://github.com/iExecBlockchainComputing/iexec-core/pull/504/commits/a05aeb4c01690740aba426c0841c1f9ad2d4b1a9).

src/main/java/com/iexec/core/replicate/ReplicateModel.java
```diff
@@ -38,7 +38,7 @@ public class ReplicateModel {
     private String resultLink;
     private String chainCallbackData;
     private String contributionHash;
-    private String appStdout;
+    private String appLogs;
```
jeremyjams, line 41:
```suggestion
    private ComputeLogs appLogs;
```

src/main/java/com/iexec/core/replicate/ReplicateModel.java
```diff
@@ -38,7 +38,7 @@ public class ReplicateModel {
...
 
     public static ReplicateModel fromEntity(Replicate entity) {
         return ReplicateModel.builder()
```
jeremyjams, line 44: `ReplicateModel.builder().computeLogs(...)`

When content is added to `ReplicateModel`, we should purge details to avoid polluting presentation layer.
Something like:
```
for (statusUpate: statusUpdates) {
statusUpdate.setComputeLogs(null);
statusUpdate.setResultLink(null);
...
}
``` 
and add annotation en `ReplicateDetails` to ignore null values

Do you see my point?

src/main/java/com/iexec/core/replicate/ReplicateModel.java
mcornaton: FYI: `appLogs` is the link to the logs, not the actual logs. So it should stay as a String :)

src/main/java/com/iexec/core/replicate/ReplicateModel.java
mcornaton: The best way to do so seems to be adding a new `ReplicateStatusUpdateModel` class. Please see https://github.com/iExecBlockchainComputing/iexec-core/pull/504/commits/815a0ebe6791ab5f9c54a28b72bde30f9b981b75.

src/main/java/com/iexec/core/replicate/ReplicateModel.java
jeremyjams: Totally right! Thanks!

src/main/java/com/iexec/core/replicate/ReplicateModel.java
jeremyjams: Excellent, removed an existing `TODO` :clap: 

Do you think
```
    private Integer exitCode; //null means unset
    private String teeSessionGenerationError; // null means unset
```
Can be moved to replicateModel
```
ReplicateStatusUpdateModel.exitCode -> ReplicateModel.appExitCode (like appLogs)
ReplicateStatusUpdateModel.teeSessionGenerationError -> ReplicateModel.teeSessionGenerationError
```
?

mcornaton: I have wondered about that. `teeSessionGenerationError` would be easy to be moved to `ReplicateModel` as we are sure there's only one.
Regarding `exitCode`, are we sure we will not want to put pre/post-compute exit codes in our models?

jeremyjams: Aren't pre & post exit codes useless today since errors are already covered by now business exceptions?

mcornaton: I think so, yes. Let's update our code to move `exitCode` and `teeSessionGenerationError` upwards :)

mcornaton: Please see https://github.com/iExecBlockchainComputing/iexec-core/pull/504/commits/330a91fc9b1c23feeefcfb3317b68bf3906e0f83.

src/main/java/com/iexec/core/replicate/ReplicateModel.java
```diff
@@ -33,22 +36,45 @@ public class ReplicateModel {
...
     public static ReplicateModel fromEntity(Replicate entity) {
+        final List<ReplicateStatusUpdateModel> statusUpdateList = new ArrayList<>();
+
+        Integer appExitCode = null;
+        String teeSessionGenerationError = null;
+        for (ReplicateStatusUpdate replicateStatusUpdate : entity.getStatusUpdateList()) {
+            statusUpdateList.add(ReplicateStatusUpdateModel.fromEntity(replicateStatusUpdate));
+            ReplicateStatusDetails details = replicateStatusUpdate.getDetails();
+            if (details != null) {
+                if (appExitCode == null) {
```
jeremyjams, line 56: A little awkward, but I think I see the issue you have
Maybe something like
```
if(this.exitCode != null && details.exitCode != null) // trying to write twice
  throw new MutipleOccurencesOfFielNotAllowed()
}
```
?

src/main/java/com/iexec/core/replicate/ReplicateModel.java
jeremyjams: Thanks!

src/main/java/com/iexec/core/replicate/ReplicateModel.java
mcornaton: Either we check `appExitCode` is null and we can assign at each iteration whatever the value of `details.getExitCode()` is, either we check `details.getExitCode()` is non-null and we assign only then. That's why this loop looks strange.

Your proposal is great and should be introduced but it does not solve my initial issue.

src/main/java/com/iexec/core/logs/ComputeLogsService.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final ComputeLogsRepository computeLogsRepository;
+
+    public ComputeLogsService(ComputeLogsRepository computeLogsRepository) {
+        this.computeLogsRepository = computeLogsRepository;
+    }
+
+    public void addComputeLogs(String chainTaskId, ComputeLogs computeLogs) {
+        TaskLogs taskLogs = getTaskLogs(chainTaskId).orElse(new TaskLogs(chainTaskId));
+        if (taskLogs.containsWalletAddress(computeLogs.getWalletAddress())) {
```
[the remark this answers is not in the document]
mcornaton: Right, see https://github.com/iExecBlockchainComputing/iexec-core/pull/504/commits/905c4a7abd32acbda4c2dd46034a49f3419bfe07.

src/main/java/com/iexec/core/replicate/ReplicateModel.java
```diff
@@ -33,22 +36,45 @@ public class ReplicateModel {
     private String chainTaskId;
     private String walletAddress;
     private ReplicateStatus currentStatus;
-    //TODO: Move/extract details here instead of encapsulating them within status updates
-    private List<ReplicateStatusUpdate> statusUpdateList;
+    private List<ReplicateStatusUpdateModel> statusUpdateList;
```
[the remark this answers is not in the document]
mcornaton: See https://github.com/iExecBlockchainComputing/iexec-core/pull/504/commits/48e0118134e4190b3e229f3c4e88ddfb2fdbb4ab.

src/main/java/com/iexec/core/logs/TaskLogs.java
```diff
@@ -49,21 +50,21 @@ public class TaskStdout {
...
         this.chainTaskId = chainTaskId;
-        this.replicateStdoutList = replicateStdoutList;
+        this.computeLogsList = computeLogsList;
     }
 
     public boolean containsWalletAddress(String walletAddress) {
-        return replicateStdoutList.stream().anyMatch(
-            replicateStdout -> replicateStdout.getWalletAddress().equals(walletAddress)
+        return computeLogsList.stream().anyMatch(
+            replicateLog -> replicateLog.getWalletAddress().equals(walletAddress)
```
jeremyjams, line 67:
```suggestion
            computeLog -> computeLog.getWalletAddress().equals(walletAddress)
```

src/test/java/com/iexec/core/logs/TaskComputeLogsServiceTests.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2020 IEXEC BLOCKCHAIN TECH
```
jeremyjams, line 2:
```suggestion
 * Copyright 2022 IEXEC BLOCKCHAIN TECH
```

src/main/java/com/iexec/core/logs/TaskLogs.java
mcornaton: Right, thanks. Fixed!

src/main/java/com/iexec/core/logs/ComputeLogsRepository.java
```diff
@@ -14,20 +14,20 @@
...
+package com.iexec.core.logs;
 
 import java.util.List;
 import java.util.Optional;
 
 import org.springframework.data.mongodb.repository.MongoRepository;
 import org.springframework.data.mongodb.repository.Query;
 
-public interface StdoutRepository extends MongoRepository<TaskStdout, String> {
+public interface ComputeLogsRepository extends MongoRepository<TaskLogs, String> {
```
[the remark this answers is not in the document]
mcornaton: It's both :grin: 
But I think we can safely say it stores `TaskLogs`. Let's rename it!

jeremyjams (approved):

PR #519, opened 2022-08-12
Run Gramine task on worker

mcornaton:


src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -54,21 +56,24 @@ class TaskUpdateManager  {
...
 
     public TaskUpdateManager(TaskService taskService,
                              IexecHubService iexecHubService,
                              ReplicatesService replicatesService,
                              ApplicationEventPublisher applicationEventPublisher,
                              WorkerService workerService,
                              BlockchainAdapterService blockchainAdapterService,
-                             SmsService smsService) {
+                             SmsService smsService,
+                             SmsClientProvider smsClientProvider) {
```
jeremyjams, line 68: I find strange at that level to to inject twice something related to SMSs. Maybe you could relocate `SmsClientProvider`  part in `SmsService` then expose a new method from there `SmsService#isClientReady(String chainTaskId)` (?)

mcornaton: Good idea, please see https://github.com/iExecBlockchainComputing/iexec-core/pull/519/commits/d92a8733145de56e753d930854748ac0524aa300.

jeremyjams (approved):
Thanks :)

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -202,6 +202,13 @@ void received2Initializing(Task task) {
             return;
         }
 
+        if (task.isTeeTask() && !smsService.isSmsClientReady(task.getChainDealId(), task.getChainTaskId())) {
+            log.error("Couldn't get SmsClient [chainTaskId: {}]", task.getChainTaskId());
+            updateTaskStatusAndSave(task, INITIALIZE_FAILED);
+            updateTaskStatusAndSave(task, FAILED);
```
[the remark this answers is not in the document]
mcornaton: As per our internal chat: let's stay as is. If we don't set task status for all failing cases, we'll get back to there later and if we should fail, we'll fail.

PR #521, opened 2022-08-31
Refactor TEE configuration

mcornaton:


jeremyjams (approved):

src/main/java/com/iexec/core/sms/SmsService.java
```diff
@@ -33,21 +38,64 @@
...
+        this.iexecHubService = iexecHubService;
     }
 
+    /**
+     * Checks the following conditions:
+     * <ul>
+     *     <li>Given deal exists on-chain;</li>
+     *     <li>The {@link SmsClient} can be created, based on the on-chain deal definition;</li>
+     *     <li>The targeted SMS is configured to run with the task's TEE enclave provider.</li>
+     * </ul>
```
jbern0rd, line 54: FYI, and it's an open discussion.
I try to follow recommended way to write javadoc comments
See https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#examples for lists.

jbern0rd (approved):

src/main/java/com/iexec/core/sms/SmsService.java
mcornaton: Regarding the closing tags: I usually prefer to include them as they are default in my IDE. I also find it clearer for someone who's used to HTML reading/writing. [Examples in the Spring Boot repo](https://github.com/search?q=%3C%2Fli%3E+repo%3Aspring-projects%2Fspring-boot+extension%3Ajava&type=Code&ref=advsearch&l=&l=).
Regarding the indentation: I also find it clearer, as in any language, to indent different levels.

We should define some good practices in our team, write them somewhere and stick to them :)

PR #535, opened 2022-12-20
Backmerge/7.2.2

jbern0rd:


src/main/java/com/iexec/core/configuration/PublicConfigurationService.java
```diff
@@ -0,0 +1,98 @@
...
+        // TODO: would be great to put this in Common
+        // (a simple `@ToString` would be sufficient)
+        final String publicConfigurationAsString = String.join("\n",
+                publicConfiguration.getWorkerPoolAddress(),
+                publicConfiguration.getBlockchainAdapterUrl(),
+                publicConfiguration.getSchedulerPublicAddress(),
+                publicConfiguration.getResultRepositoryURL(),
+                publicConfiguration.getAskForReplicatePeriod() + "",
+                publicConfiguration.getRequiredWorkerVersion()
+        );
```
mcornaton, line 80: Shouldn't we update it right now with `publicConfiguration.toString()`? Or is it too risky?

mcornaton: We'll do it in a later PR :)

mcornaton (approved):

src/main/java/com/iexec/core/configuration/PublicConfigurationService.java
jbern0rd: I'm more enclined to do this in a next PR to go fast here.

PR #541, opened 2023-01-04
Bugfix/code smells in tests

jbern0rd:


jeremyjams (approved):

src/test/java/com/iexec/core/worker/WorkerServiceTests.java
```diff
@@ -65,8 +65,8 @@ void shouldGetWorker() {
 
         when(workerRepository.findByWalletAddress(walletAddress)).thenReturn(Optional.of(existingWorker));
         Optional<Worker> foundWorker = workerService.getWorker(walletAddress);
-        assertThat(foundWorker.isPresent()).isTrue();
-        assertThat(foundWorker.get()).isEqualTo(existingWorker);
+        assertThat(foundWorker).isPresent();
+        assertThat(foundWorker).contains(existingWorker);
```
mcornaton, line 69:
```suggestion
        assertThat(foundWorker).isPresent().contains(existingWorker);
```

mcornaton (approved):

src/test/java/com/iexec/core/worker/WorkerServiceTests.java
jbern0rd: Why do you suggest this ? If it is not present, it will not contain the expected result.
Both old checks can be merged by a single one.

mcornaton: That's a code smell: https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=541&id=com.iexec.core%3Aiexec-core&open=AYV96ku70Yjce71DVeyo
I guess that, if the first one throws, then the second one is not executed.

jbern0rd: Code smell fixed in 1520c09, I missed it in your suggestion.
I removed the check on `isPresent`.

PR #543, opened 2023-01-11
Bugfix/task added twice

mcornaton:


jbern0rd (approved):

jbern0rd (approved):

src/test/java/com/iexec/core/task/TaskServiceRealRepositoryTest.java
```diff
@@ -0,0 +1,105 @@
...
+                    2, maxExecutionTime, "0x0", contributionDeadline, finalDeadline)));
+        }
+
+        // Let's wait for the `taskService.addTask` to complete.
+        executions.forEach(execution -> {
+            try {
+                execution.get(1, TimeUnit.MINUTES);
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof DuplicateKeyException) {
+                    fail("Task has been added twice. Should not happen!");
```
jeremyjams, line 95: I'm not sure this can happen :thinking:

mcornaton: You're right. This should not happen anymore. However, I'd like to keep it there. This would ensure that, if we ever refactor `TaskService#addTask`, we still handle such a case.

jbern0rd: I agree. Shouldn't we try to check the `addTask` method returns the correct value as well ?

jeremyjams (approved):
:+1:

src/test/java/com/iexec/core/task/TaskServiceRealRepositoryTest.java
mcornaton: Please see this other PR: https://github.com/iExecBlockchainComputing/iexec-core/pull/547

PR #547, opened 2023-01-30
Check task is correctly added on parallel `TaskService#addTask` calls

mcornaton:


jbern0rd (approved):
Thanks

src/test/java/com/iexec/core/task/TaskServiceRealRepositoryTest.java
```diff
@@ -86,20 +84,32 @@ void shouldAddTaskASingleTime() {
...
                 throw new RuntimeException(e);
             }
-        });
+        }).collect(Collectors.toList());
+
+        // Check one execution has added the task,
+        // while the other one has failed.
+        final Optional<Task> task1 = results.get(0);
+        final Optional<Task> task2 = results.get(1);
+        assertThat(task1.isEmpty() || task2.isEmpty()).isTrue();
```
jeremyjams, line 105: Aren't you looking for a XOR `^` instead?

mcornaton: You're right! I have looked for `assertThatExactlyOneIsTrue` or some similar method, but couldn't get any result. Thanks for the XOR idea, it's a good amelioration.
Please see https://github.com/iExecBlockchainComputing/iexec-core/pull/547/commits/467da7da0063895bd69eee35067341714b21d409.

mcornaton: Based on a suggestion from @jbern0rd, I've refactored this part. We can now add more than 2 tasks at the same time. This helps to prove the robustness of our code. Please see https://github.com/iExecBlockchainComputing/iexec-core/pull/547/commits/929b3fc1e85380142c43078fbec00e670360eb21.

jeremyjams (approved):
Thanks :+1:

PR #551, opened 2023-02-15
Hotfix/7.3.1

jbern0rd:


src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -175,7 +191,36 @@ void replayDealEvent() {
...
+        return filter;
+    }
+
+    Optional<DealEvent> schedulerNoticeToDealEvent(IexecHubContract.SchedulerNoticeEventResponse schedulerNotice) {
+        dealEventsCount++;
+        BigInteger noticeBlockNumber = schedulerNotice.log.getBlockNumber();
+        if (latestBlockNumberWithDeal.compareTo(noticeBlockNumber) < 0) {
+            latestBlockNumberWithDeal = noticeBlockNumber;
+        }
+        log.info("Received {} deal events notifications since scheduler startup", dealEventsCount);
```
jeremyjams, line 219: `Received new deal [<dealEventsCount>,<dealId>]`

src/test/java/com/iexec/core/metric/MetricServiceTests.java
```diff
@@ -56,6 +60,8 @@ void shouldGetPlatformMetrics() {
         when(workerService.getAliveAvailableGpu()).thenReturn(1);
         when(taskService.findByCurrentStatus(TaskStatus.COMPLETED))
                 .thenReturn(List.of());
+        when(dealWatcherService.getDealEventsCount()).thenReturn(0);
```
jeremyjams, line 63: Maybe some non zero values?

jbern0rd: Sure, see fcd0fd8

src/main/java/com/iexec/core/chain/DealWatcherService.java
jbern0rd: Logging event updated, see 5c2ba25

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -175,7 +190,36 @@ void replayDealEvent() {
...
+        dealEventsCount++;
+        BigInteger noticeBlockNumber = schedulerNotice.log.getBlockNumber();
+        if (latestBlockNumberWithDeal.compareTo(noticeBlockNumber) < 0) {
+            latestBlockNumberWithDeal = noticeBlockNumber;
+        }
+        log.info("Received new deal [dealEventsCount:{}, chainDealId:{}]", dealEventsCount, BytesUtils.bytesToString(schedulerNotice.dealid));
+        if (schedulerNotice.workerpool.equalsIgnoreCase(chainConfig.getPoolAddress())) {
+            return Optional.of(new DealEvent(schedulerNotice));
+        }
+        log.warn("This deal event should not have been received");
```
mcornaton, line 222: Could we log the block number as well? With multi-threading, this log line could be way away from the `Received new deal` line.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## [[7.3.1]](https://github.com/iExecBlockchainComputing/iexec-core/releases/tag/v7.3.1) 2023-02-17
+
+* Improve filter to receive deal events addressed to the current _iExec Core Scheduler_. Improve `IEXEC_START_BLOCK_NUMBER` documentation.
```
jeremyjams, line 7:
```suggestion
* Subscribe only to deal events targeting a specific workerpool. Improve `IEXEC_START_BLOCK_NUMBER` documentation.
```

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -175,7 +190,36 @@ void replayDealEvent() {
...
+    }
+
+    Optional<DealEvent> schedulerNoticeToDealEvent(IexecHubContract.SchedulerNoticeEventResponse schedulerNotice) {
+        dealEventsCount++;
+        BigInteger noticeBlockNumber = schedulerNotice.log.getBlockNumber();
+        if (latestBlockNumberWithDeal.compareTo(noticeBlockNumber) < 0) {
+            latestBlockNumberWithDeal = noticeBlockNumber;
+        }
+        log.info("Received new deal [dealEventsCount:{}, chainDealId:{}]", dealEventsCount, BytesUtils.bytesToString(schedulerNotice.dealid));
+        if (schedulerNotice.workerpool.equalsIgnoreCase(chainConfig.getPoolAddress())) {
```
mcornaton, line 219: I usually return the error case instead of the successful case in an `if` block. I find it clearer as this is what I'm expecting to read. Something like:
```java
if (!schedulerNotice.workerpool.equalsIgnoreCase(chainConfig.getPoolAddress())) {
    log.warn(...);
    return Optional.empty();
}
return Optional.of(...);
```

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -95,11 +111,10 @@ private void onDealEvent(DealEvent dealEvent) {
         BigInteger dealBlock = dealEvent.getBlockNumber();
         log.info("Received deal [dealId:{}, block:{}]", dealId,
                 dealBlock);
-        if (dealBlock == null || dealBlock.equals(BigInteger.ZERO)){
+        if (dealBlock.equals(BigInteger.ZERO)) {
```
jeremyjams, line 114: Why can't it be null?

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -175,7 +190,36 @@ void replayDealEvent() {
...
+        }
+        EthFilter filter = new EthFilter(fromBlock, toBlock, chainConfig.getHubAddress());
+        BigInteger poolAddressBigInt = Numeric.toBigInt(chainConfig.getPoolAddress());
+        filter.addSingleTopic(EventEncoder.encode(SCHEDULERNOTICE_EVENT));
+        filter.addSingleTopic(Numeric.toHexStringWithPrefixZeroPadded(poolAddressBigInt, 64));
+        return filter;
+    }
+
+    Optional<DealEvent> schedulerNoticeToDealEvent(IexecHubContract.SchedulerNoticeEventResponse schedulerNotice) {
+        dealEventsCount++;
```
jeremyjams, line 213: Isn't your counter is incremented twice (standard + replay)?

src/main/java/com/iexec/core/chain/DealWatcherService.java
jeremyjams: I personally think error should be the default, and the good path a chance.
(= I like current implem* :) )

mcornaton: Well, maybe both of our opinions are right here. I mean, when there are multiple failure cases and a single success, the success usually is the last line of the method. But when there are multiple success success and a single "default" failure case, then this failure case is the last line.
Here, a single success case and a single failure case, both implementations would work.

src/main/java/com/iexec/core/chain/DealWatcherService.java
jbern0rd: That is the goal. I want to count all received deal events, even if we receive an event for an already notified event.
It was hard to show in both PRs, but this allowed me a TDD approach and to check the workerpool only receive its specific events.
In the future, when we rework the DealWatcherService, the `dealEventsCount` value will converge to the `receivedDealsCount`.

README.md
```diff
@@ -57,6 +57,12 @@ You can configure the _iExec Core Scheduler_ with the following properties:
 | `IEXEC_LOGS_PURGE_RATE_IN_DAYS` | Interval in days between 2 executions of the purge mechanism. | Positive integer | `1` |
 | `IEXEC_LOGS_AVAILABILITY_PERIOD_IN_DAYS` | Number of days to keep logs of past tasks. | Positive integer | `3` |
 
+> If it is not the first startup of the _iExec Core Scheduler_ and if it received deals previously,
+> the _MongoDB_ instance will contain a __configuration Collection__ in the __iexec Database__.
+> The value stored in this document takes the precedence over the `IEXEC_START_BLOCK_NUMBER` configuration parameter.
+> To enforce deal observation starting from the `IEXEC_START_BLOCK_NUMBER` value, the aforementioned document has to be deleted in the _MongoDB_.
+> All deals prior to the `IEXEC_START_BLOCK_NUMBER` will then be ignored.
```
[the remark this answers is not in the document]
jbern0rd: I tried to see if it allowed to put emphasis on this information where `BigInteger.ZERO` is enforced instead of `null`.
This allowed to create an immutable non-null `DealEvent` and to remove the call to `setBlockNumber`.

src/main/java/com/iexec/core/chain/DealWatcherService.java
jbern0rd: This is directly related to commit 77b6300.
The `BigInteger.ZERO` value is enforced instead of null.
This allowed to remove the call to `setBlockNumber` and to work with immutable `DealEvent` instances.

CHANGELOG.md
jbern0rd: Sure, please see 48e6996

README.md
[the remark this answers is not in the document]
jbern0rd: I tried this to put emphasis on this, but it is not that great.
Quote has been removed in 48e6996

src/main/java/com/iexec/core/chain/DealWatcherService.java
jbern0rd: Sure, please see 2486787

src/main/java/com/iexec/core/chain/DealWatcherService.java
```diff
@@ -175,7 +190,36 @@ void replayDealEvent() {
...
+                .map(this::schedulerNoticeToDealEvent)
                 .subscribe(dealEvent -> dealEvent.ifPresent(this::onDealEvent));
     }
+
+    EthFilter createDealEventFilter(BigInteger from, BigInteger to) {
+        DefaultBlockParameter fromBlock = DefaultBlockParameter.valueOf(from);
+        DefaultBlockParameter toBlock = DefaultBlockParameterName.LATEST;
+        if (to != null) {
+            toBlock = DefaultBlockParameter.valueOf(to);
+        }
```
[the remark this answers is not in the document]
jbern0rd: see 9e5da04 and a45ae91

src/main/java/com/iexec/core/chain/DealWatcherService.java
jeremyjams: I see, thank you!

jeremyjams (approved):

mcornaton (approved):

PR #554, opened 2023-02-21
Feature/cache jwt until expiration

jbern0rd:


src/main/java/com/iexec/core/security/ChallengeService.java
```diff
@@ -16,30 +16,23 @@
...
+    private final ConcurrentHashMap<String, String> challengesMap = new ConcurrentHashMap<>();
 
-    ChallengeService() {
-        this.challengeMap = ExpiringMap.builder()
-                .expiration(60, TimeUnit.MINUTES)
-                .expirationPolicy(ExpirationPolicy.CREATED)
-                .build();
+    public String getChallenge(String workerWallet) {
+        challengesMap.computeIfAbsent(workerWallet, wallet -> RandomStringUtils.randomAlphabetic(10));
+        return challengesMap.get(workerWallet);
```
mcornaton, line 31:
```suggestion
        return challengesMap.computeIfAbsent(workerWallet, wallet -> RandomStringUtils.randomAlphabetic(10));
```
`computeIfAbsent` returns the current or newly-computed value

mcornaton, line 31:
```suggestion
        return challengesMap.computeIfAbsent(workerWallet, wallet -> RandomStringUtils.randomAlphabetic(10));
```
`computeIfAbsent` returns the current or newly-computed value

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -42,13 +42,25 @@ public JwtTokenProvider(ChallengeService challengeService) {
...
+            Date now = new Date();
+            return Jwts.builder()
+                    .setAudience(address)
+                    .setIssuedAt(now)
+                    .setExpiration(new Date(now.getTime() + TOKEN_VALIDITY_DURATION))
+                    .setSubject(challengeService.getChallenge(address))
+                    .signWith(SignatureAlgorithm.HS256, secretKey)
+                    .compact();
+        });
+        return jwTokensMap.get(walletAddress);
```
mcornaton, line 63:
```suggestion
        return jwTokensMap.computeIfAbsent(walletAddress, address -> {
            Date now = new Date();
            return Jwts.builder()
                    .setAudience(address)
                    .setIssuedAt(now)
                    .setExpiration(new Date(now.getTime() + TOKEN_VALIDITY_DURATION))
                    .setSubject(challengeService.getChallenge(address))
                    .signWith(SignatureAlgorithm.HS256, secretKey)
                    .compact();
        });
```
Same as before :)

mcornaton, line 63:
```suggestion
        return jwTokensMap.computeIfAbsent(walletAddress, address -> {
            Date now = new Date();
            return Jwts.builder()
                    .setAudience(address)
                    .setIssuedAt(now)
                    .setExpiration(new Date(now.getTime() + TOKEN_VALIDITY_DURATION))
                    .setSubject(challengeService.getChallenge(address))
                    .signWith(SignatureAlgorithm.HS256, secretKey)
                    .compact();
        });
```
Same as before :)

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -42,13 +42,25 @@ public JwtTokenProvider(ChallengeService challengeService) {
...
     }
 
+    /**
+     * Creates a signed JWT with expiration date for a given ethereum address.
+     * <p>
+     * The token is cached until it expires.
+     * @param walletAddress worker address for which the token is created
+     * @return A signed JWT for a given ethereum address
+     */
     public String createToken(String walletAddress) {
```
mcornaton, line 52: Shouldn't we check the expiration of the token there as well?

src/test/java/com/iexec/core/security/JwtTokenProviderTests.java
```diff
@@ -73,6 +78,21 @@ void isValidTokenTrue() {
         assertThat(isValidToken).isTrue();
     }
 
+    @Test
+    void isValidTokenFalseSinceExpired() {
+        when(challengeService.getChallenge(WALLET_ADDRESS)).thenReturn(" challenge");
+        String token = jwtTokenProvider.createToken(WALLET_ADDRESS);
+        new DefaultClaims().setAudience(WALLET_ADDRESS);
+        try (MockedStatic<Jwts> mockedStatic = mockStatic(Jwts.class)) {
```
mcornaton, line 86: Are we forced to mock this static class? Can't we forge an expired token?

src/main/java/com/iexec/core/security/ChallengeService.java
```diff
@@ -16,30 +16,23 @@
...
-    private final ExpiringMap<String, String> challengeMap;
+    private final ConcurrentHashMap<String, String> challengesMap = new ConcurrentHashMap<>();
 
-    ChallengeService() {
-        this.challengeMap = ExpiringMap.builder()
-                .expiration(60, TimeUnit.MINUTES)
-                .expirationPolicy(ExpirationPolicy.CREATED)
-                .build();
+    public String getChallenge(String workerWallet) {
+        challengesMap.computeIfAbsent(workerWallet, wallet -> RandomStringUtils.randomAlphabetic(10));
```
mcornaton: There's a security hotspot there: https://sonarcloud.io/project/security_hotspots?id=com.iexec.core%3Aiexec-core&pullRequest=554&resolved=false&types=SECURITY_HOTSPOT&tab=code

src/test/java/com/iexec/core/security/JwtTokenProviderTests.java
jbern0rd: not really easy, need to inject the encryption key and the forged token in JwtTokenProvider

src/main/java/com/iexec/core/security/JwtTokenProvider.java
jbern0rd: Fixed in e2e5ad8

src/test/java/com/iexec/core/security/JwtTokenProviderTests.java
mcornaton: Alright. Let's keep it as is then :)

src/main/java/com/iexec/core/security/ChallengeService.java
jbern0rd: Yes, as said it is a weak cryptography hotspot.
We either decide it is not a problem or we use a crypto mechanism such as UUID which does not trigger a hotspot.

src/main/java/com/iexec/core/security/JwtTokenProvider.java
jbern0rd: The `createToken` is only called from `WorkerController#login` which should almost only happen when a worker receives an `UNAUTHORIZED` HTTP status.
A worker receives such a status when a token is not valid (not signed, expired or bad claims).
This is true for the `/replicates/available`, `/replicates/interrupted`, `/replicates/{chainTaskId}/status`, `/workers/ping`, `/workers/register`, `/workers/computing` endpoints which should be used more often than the `/worker/login` one.

I think it would be better to keep the current algorithm here:
- The worker provides a token on an aforementioned endpoint
- When the token is expired
  - it is purged from its cache (as well as the challenge)
  - the worker receives HTTP status UNAUTHORIZED
  - the worker launches the login procedure in `LoginService` (create new challenge, create new token)
- The token is valid on next REST call requiring it

With this flow, almost no call to `/workers/login` should be made with an expired token still existing in the cache.

To go your way would require to modify `createToken`:
- to check if a value/token exists for the provided address
- if it exists, to check if it is valid
  - when valid, return the token
  - if invalid, challenge and token will be purged from maps
  - call to `/worker/login` will fail in worker

It seems to be a more complex implementation which only allow to gain one HTTP request.
I am not sure the gain is really worth the investment.
What do you think ?

mcornaton: It doesn't seem it would need a lot of investment. I guess something like that would work:
```java
public String createToken(String walletAddress) {
    final String token = jwTokensMap.get(walletAddress);
    if (!StringUtils.isEmpty(token) && isValidToken(token)) {
        // token is valid, reusing it
        return token;
    }

    // token is invalid, renewing it
    challengeService.removeChallenge(walletAddress);
    return jwTokensMap.compute(walletAddress, (address, oldToken) -> {
        Date now = new Date();
        return Jwts.builder()
                .setAudience(address)
                .setIssuedAt(now)
                .setExpiration(new Date(now.getTime() + TOKEN_VALIDITY_DURATION))
                .setSubject(challengeService.getChallenge(address))
                .signWith(SignatureAlgorithm.HS256, secretKey)
                .compact();
    });
}
```

In fact, I usually find it better to avoid communications (HTTP requests) when they are not required.

However, if we decide to keep your version, we should at least fix the JavaDoc. It currently says the following:
> The token is cached until it expires.
This is not true, as this method can return an expired token that will be purged only on next authenticated call.

src/main/java/com/iexec/core/security/ChallengeService.java
jbern0rd: see d66528e

src/main/java/com/iexec/core/security/JwtTokenProvider.java
jbern0rd: see b56c839

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -50,6 +51,14 @@ public JwtTokenProvider(ChallengeService challengeService) {
      * @return A signed JWT for a given ethereum address
      */
     public String createToken(String walletAddress) {
+        final String token = jwTokensMap.get(walletAddress);
+        if (!StringUtils.isEmpty(token) && isValidToken((token))) {
+            log.info("token found");
```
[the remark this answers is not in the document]
mcornaton: Removed in [the last commi](https://github.com/iExecBlockchainComputing/iexec-core/pull/554/commits/589eca66f95d31ee58263ac5182c10ae20d8f9cc)

mcornaton (approved):
Looks great, thanks!

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -58,46 +74,50 @@ public String resolveToken(String token) {
...
-     *  3) old challenge expires
-     *  4) worker tries logging with old challenge
+    /**
+     * Checks if a JWT is valid.
+     * <p>
+     * A valid JWT must:
+     * <ul>
+     * <li>be signed with the scheduler private key
+     * <li>not be expired
+     * <li>contain valid address and challenge values respectively in audience and subject claims
```
jeremyjams, line 84: Can you remove/update these comments (mostly challenge-related)

src/main/java/com/iexec/core/worker/WorkerController.java
```diff
@@ -85,6 +85,7 @@ public ResponseEntity<String> getToken(@RequestParam(name = "walletAddress") Str
 
         String challenge = challengeService.getChallenge(walletAddress);
         byte[] hashToCheck = Hash.sha3(BytesUtils.stringToBytes(challenge));
+        challengeService.removeChallenge(walletAddress, challenge);
```
jeremyjams, line 88: Challenge must be removed after its verification, otherwise Mallory could attack Alice challenge by impersonating her while spamming /login

src/main/java/com/iexec/core/security/JwtTokenProvider.java
jbern0rd: Sure, let me know if it is better in 52483222

src/main/java/com/iexec/core/worker/WorkerController.java
jbern0rd: Fixed in 8349102

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -16,39 +16,55 @@
...
-        return Jwts.builder()
-                .setAudience(walletAddress)
-                .setIssuedAt(new Date())
-                .setSubject(challengeService.getChallenge(walletAddress))
-                .signWith(SignatureAlgorithm.HS256, secretKey)
-                .compact();
+    /**
+     * Creates a signed JWT with expiration date for a given ethereum address.
+     * <p>
+     * The token is cached until it expires.
```
jeremyjams, line 49:
```suggestion
     * The token is cached. It might be pruned in best effort mode by other processes founding that token is expired.
```

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -58,46 +74,50 @@ public String resolveToken(String token) {
...
-     *  3) old challenge expires
-     *  4) worker tries logging with old challenge
+    /**
+     * Checks if a JWT is valid.
+     * <p>
+     * A valid JWT must:
+     * <ul>
+     * <li>be signed with the scheduler private key
+     * <li>not be expired
+     * <li>contain th scheduler application ID in the audience claim
```
jeremyjams, line 84:
```suggestion
     * <li>contain the scheduler application ID in the audience claim
```

src/main/java/com/iexec/core/security/JwtTokenProvider.java
```diff
@@ -58,46 +74,50 @@ public String resolveToken(String token) {
...
+     * <p>
+     * An invalid JWT will return an UNAUTHORIZED status and require to perform a full authentication loop
+     * with a new signed challenge (get new challenge -> sign challenge -> check signed challenge -> get or create JWT).
+     * <p>
+     * If the JWT was expired, the cache will have been purged and a new JWT will be generated.
```
jeremyjams, line 89:
```suggestion
     * If the JWT has expired, the cache will be purged and a new JWT will be generated.
```

src/test/java/com/iexec/core/security/JwtTokenProviderTests.java
```diff
@@ -67,24 +88,59 @@ void shouldNotResolveTokenSinceNullOne() {
...
+    void isValidTokenFalseSinceNotSigned() {
+        String token = Jwts.builder()
+                .setAudience(JWT_AUDIENCE)
+                .setIssuedAt(new Date())
+                .setSubject(WALLET_ADDRESS)
+                .compact();
+        boolean isTokenValid = jwtTokenProvider.isValidToken(token);
+        assertThat(isTokenValid).isFalse();
+    }
+
```
jeremyjams, line 140:
```suggestion

//TODO: Add isValidTokenFalseSinceBadSigned

```
?

jeremyjams (approved):
Thank you!

src/main/java/com/iexec/core/security/ChallengeService.java
```diff
@@ -18,28 +18,33 @@
```
jeremyjams, line 18: Can you update the changelog please?

mcornaton (approved):

PR #558, opened 2023-03-02
Bugfix/always return notification on replicate status update

jbern0rd:


src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -124,16 +124,17 @@ public ResponseEntity<TaskNotificationType> updateReplicateStatus(
                 return replicatesService
                         .updateReplicateStatus(chainTaskId, walletAddress, statusUpdate, updateReplicateStatusArgs)
                         .map(ResponseEntity::ok)
-                        .orElse(ResponseEntity.status(HttpStatus.FORBIDDEN.value())
-                                .build());
+                        .orElse(ResponseEntity.status(HttpStatus.FORBIDDEN)
+                                .body(TaskNotificationType.PLEASE_ABORT));
```
jeremyjams, line 128: Thank you very much for that, maybe add a note here about the why?

mcornaton: I second that. It would be great to have to comment to remember later why we did this :)

mcornaton (approved):

src/main/java/com/iexec/core/replicate/ReplicatesController.java
jbern0rd: Does the last commit help ?

jbern0rd: See 57143be

jeremyjams: Perfect, thanks!

jeremyjams (approved):

PR #564, opened 2023-03-15
Always send a `TaskNotificationType` with a 2XX HTTP status code

jbern0rd:


mcornaton (approved):

src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -21,6 +21,7 @@
 import com.iexec.common.replicate.*;
 import com.iexec.core.security.JwtTokenProvider;
 import com.iexec.core.worker.WorkerService;
+import feign.FeignException;
```
jeremyjams, line 24:
```suggestion
```

jeremyjams (approved):
OK, thanks :+1:

src/main/java/com/iexec/core/replicate/ReplicatesController.java
jbern0rd: The dependency required after  @mcornaton recommended me to use a link in javadoc comment.
I could remove it by replacing {@link} with {@code} but the integration is not the same in IDE.

jeremyjams: Interesting :bulb:

PR #568, opened 2023-03-20
Fix race condition on replicate status update

mcornaton:


jbern0rd (approved):
Thanks for this and the last formatting.
I find it much more readable this way :+1:

PR #574, opened 2023-05-03
Add ContributeAndFinalize to `ReplicateWorkflow`

mcornaton:


src/main/java/com/iexec/core/workflow/ReplicateWorkflow.java
```diff
@@ -98,6 +98,10 @@ private void setDefaultWorkflowTransitions() {
         addTransition(RESULT_UPLOAD_FAILED, toList(COMPLETING));
         addTransition(RESULT_UPLOADED, toList(COMPLETING));
 
+        // contribute and finalize
+        addTransition(CONTRIBUTE_AND_FINALIZE_ONGOING, toList(CONTRIBUTE_AND_FINALIZE_FAILED, CONTRIBUTE_AND_FINALIZE_DONE));
```
jbern0rd, line 102: For this transition, the convention seems to be to declare `PASS` transition before `FAIL` transition.
Have a look at `CONTRIBUTING` or `REVEALED` for comparison

jbern0rd: I agree, this should be easier after `iexec-core-client` jar creation

jbern0rd (approved):

PR #579, opened 2023-05-12
disable contribute and finalize when callback is filled

thewhitewizard:


mcornaton (approved):
Thanks for this quick fix!

PR #601, opened 2023-06-26
Add `BlockchainConnectionHealthIndicator`

mcornaton:


thewhitewizard (approved):

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
```diff
@@ -0,0 +1,165 @@
+package com.iexec.core.chain;
```
jbern0rd, line 1: The license is missing on this new file

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    private void connectionSucceeded() {
+        if (!outOfService) {
+            if (consecutiveFailures > 0) {
+                log.info("Blockchain connection is now restored after a period of unavailability." +
+                        " [unavailabilityPeriod:{}]", pollingInterval.multipliedBy(consecutiveFailures));
+            }
+            firstFailure = null;
+        }
+        consecutiveFailures = 0;
```
jbern0rd, line 146: Could it be simplified as follow ?
```suggestion
        if (!outOfService && consecutiveFailures > 0) {
            log.info("Blockchain connection is now restored after a period of unavailability." +
                    " [unavailabilityPeriod:{}]", pollingInterval.multipliedBy(consecutiveFailures));
            firstFailure = null;
            consecutiveFailures = 0;
        }
```

src/test/java/com/iexec/core/chain/BlockchainConnectionHealthIndicatorTests.java
```diff
@@ -0,0 +1,218 @@
+package com.iexec.core.chain;
```
jbern0rd, line 1: The license is missing here :-)

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    private final int outOfServiceThreshold;
+    private final ScheduledExecutorService monitoringExecutor;
+
+    /**
+     * Current number of consecutive failures.
+     */
+    private int consecutiveFailures = 0;
+    private LocalDateTime firstFailure = null;
+    private boolean outOfService = false;
```
jbern0rd, line 39: What do you think to add getters to those fields ?
* Simplify tests
* Would allow us to expose them to the metrics endpoint (would be nice to see the graph of temporary communication loss)

src/test/java/com/iexec/core/chain/BlockchainConnectionHealthIndicatorTests.java
```diff
@@ -0,0 +1,218 @@
...
+     * in the following order:
+     * <ol>
+     *     <li>consecutiveFailures (initial state)</li>
+     *     <li>outOfService (initial state)</li>
+     *     <li>firstFailure (initial state)</li>
+     *     <li>latestBlockNumber (input)</li>
+     *     <li>consecutiveFailures (output)</li>
+     *     <li>outOfService (output)</li>
+     *     <li>firstFailure (output)</li>
+     * </ol>
```
jbern0rd, line 80: I think the javadoc best practice is more something like
```suggestion
     * <ol>
     * <li>consecutiveFailures (initial state)
     * <li>outOfService (initial state
     * <li>firstFailure (initial state)
     * <li>latestBlockNumber (input)
     * <li>consecutiveFailures (output)
     * <li>outOfService (output)
     * <li>firstFailure (output)
```

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
mcornaton: This version wouldn't work. `consecutiveFailures` has to be reset every time the connection succeeds. However, the following would work:
```suggestion
        if (!outOfService && consecutiveFailures > 0) {
            log.info("Blockchain connection is now restored after a period of unavailability." +
                    " [unavailabilityPeriod:{}]", pollingInterval.multipliedBy(consecutiveFailures));
            firstFailure = null;
        }
        consecutiveFailures = 0;
```
I'll apply it as it simplifies our code. Thanks for the suggestion!

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
mcornaton: I'm unsure about the second point. Do we want the information in the `/metrics` endpoint  while we have it in the `/actuator/health/blockchainConnection` endpoint?
If we don't, I'd prefer to keep it as is.

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
mcornaton: Please see https://github.com/iExecBlockchainComputing/iexec-core/pull/601/commits/befbca97c9f719e79b6f324eb72e169bf55c200b.

src/test/java/com/iexec/core/chain/BlockchainConnectionHealthIndicatorTests.java
mcornaton: Please see https://github.com/iExecBlockchainComputing/iexec-core/pull/601/commits/1b28e96e5859a2b84300736dde6fd7faaf416bef.

jbern0rd: Thanks :pray:

src/main/java/com/iexec/core/chain/BlockchainConnectionHealthIndicator.java
jbern0rd: The goal of the metric is to send data to a metric aggregator.
The health endpoint is different and should not be scraped by a monitoring endpoint.
If we want to follow communication loss without out of service, we need to do it through a metric imho.
I'd be happy to have a live discussion on this if you want to.

mcornaton: I see your point. Here's the commit: https://github.com/iExecBlockchainComputing/iexec-core/pull/601/commits/c49ae6ee11655cc03fdce957687146867d17e4ae

jbern0rd (approved):

PR #604, opened 2023-06-30
Feature/block communications while chain down

mcornaton:


src/main/java/com/iexec/core/replicate/ReplicatesController.java
```diff
@@ -61,6 +65,12 @@ public ResponseEntity<ReplicateTaskSummary> getAvailableReplicateTaskSummary(
             return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
         }
 
+        if (!blockchainConnectionHealthIndicator.isUp()) {
```
jbern0rd, line 68: Do we want to block only when blockchain connection is lost or when overall health status is down as well ?

mcornaton: It is hard to tell now which alternative suits us best. I guess we'll have to wait some time before having enough information. Let's postpone this discussion!

jbern0rd: ok :+1: 

Second question, shouldn't we execute this check first ?
When reviewing, it feels weird to me to check authorization before checking OUT-OF-SERVICE.
It would be more logical and would allow us to answer OUT-OF-SERVICE on database connection loss as well in the future.

jbern0rd: Sure, we can talk about this at a later date. :+1:

jbern0rd (approved):

PR #612, opened 2023-08-28
Feature/gradle 8.2.1

jbern0rd:


build.gradle
```diff
@@ -91,7 +91,9 @@ dependencies {
     // vavr
     implementation 'io.vavr:vavr:0.10.4'
 
+    // tests
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
+    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
```
mcornaton, line 96: Should we stay consistent with the other lines?
```suggestion
    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```

mcornaton (approved):

mcornaton (approved):

build.gradle
jbern0rd: We follow gradle upgrade guide here
https://docs.gradle.org/current/userguide/upgrading_version_8.html#manually_declaring_dependencies

PR #619, opened 2023-09-27
Add `isEligibleToContributeAndFinalize` method to `Task`

mcornaton:


src/main/java/com/iexec/core/task/Task.java
```diff
@@ -178,4 +178,19 @@ public boolean inCompletionPhase() {
...
+     * <ul>
+     *     <li>It is a TEE task
+     *     <li>Its trust is 1
+     * </ul>
+     *
+     * @return {@literal true} if eligible, {@literal false} otherwise.
+     */
+    public boolean isEligibleToContributeAndFinalize() {
+        return isTeeTask()
+                && trust == 1;
```
jbern0rd, line 194: There is a difference with iexec-commons-poco where we check if there is a callback as well.

jbern0rd: Wouldn't it be better to implement the same check ?

src/main/java/com/iexec/core/task/Task.java
```diff
@@ -178,4 +178,5 @@ public boolean inCompletionPhase() {
     public boolean isTeeTask() {
         return TeeUtils.isTeeTag(getTag());
     }
+
```
mcornaton, line 181:
```suggestion
```

jbern0rd (approved):
LGTM :rocket:

PR #626, opened 2023-11-07
Move first DTO classes to `iexec-core-library` sub-project

jbern0rd:


iexec-core-library/src/test/java/com/iexec/core/task/TaskStatusTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+    // region contribution phase
+    @ParameterizedTest
+    @EnumSource(value = TaskStatus.class, names = {"INITIALIZED", "RUNNING"})
+    void isInContributionPhase(TaskStatus taskStatus) {
+        assertTrue(TaskStatus.isInContributionPhase(taskStatus));
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = TaskStatus.class, names = {"INITIALIZED", "RUNNING"}, mode = EnumSource.Mode.EXCLUDE)
```
mcornaton, line 35: For clarity purpose, should we reorder params here? I'd like to see `mode` before `names`. That would make it easier to read: "Ok, we'll use TaskStatus values. The following values are excluded: INITIALIZED, RUNNING. Got it!" instead of: "Ok, we'll use TaskStatus values. These are INITIALIZED, RUNNING. Got it! Oh, no, they are excluded."
Not a big deal, but would be great IMHO :)

mcornaton (approved):

iexec-core-library/src/test/java/com/iexec/core/task/TaskStatusTests.java
jbern0rd: see f433fbb

PR #632, opened 2023-11-15
Feature/add prometheus endpoint

thewhitewizard:


src/main/resources/application.yml
```diff
@@ -79,7 +80,7 @@ management:
         # health, info, httptrace
         # or:
         # *
-        include: ${IEXEC_CORE_MANAGEMENT_ACTUATORS:health, info}
+        include: ${IEXEC_CORE_MANAGEMENT_ACTUATORS:health, info, metrics}
```
mcornaton, line 83: Do we need to expose `metrics` endpoint as a default? I'd prefer it to be exposed only if the admin wants to.

src/main/java/com/iexec/core/metric/MetricService.java
```diff
@@ -37,6 +53,23 @@ public MetricService(DealWatcherService dealWatcherService,
...
+        aliveAvailableCpuCount = Metrics.gauge("iexec.core.cpu.available", new AtomicInteger(workerService.getAliveAvailableCpu()));
+        completedTasksCount = Metrics.gauge("iexec.core.tasks.completed", new AtomicInteger(taskService.findByCurrentStatus(TaskStatus.COMPLETED).size()));
+        dealEventsCount = Metrics.gauge("iexec.core.deals.events", new AtomicLong(dealWatcherService.getDealEventsCount()));
+        dealsCount = Metrics.gauge("iexec.core.deals", new AtomicLong(dealWatcherService.getDealsCount()));
+        replayDealsCount = Metrics.gauge("iexec.core.deals.replay", new AtomicLong(dealWatcherService.getReplayDealsCount()));
+        if (null != dealWatcherService.getLatestBlockNumberWithDeal()) {
+            latestBlockNumberWithDeal = Metrics.gauge("iexec.core.deals.last.block", new AtomicLong(dealWatcherService.getLatestBlockNumberWithDeal().longValue()));
+        } else {
+            latestBlockNumberWithDeal = Metrics.gauge("iexec.core.deals.last.block", new AtomicLong(-1L));
+        }
```
mcornaton, line 70: We could avoid a double call to `getLatestBlockNumberWithDeal`:
```suggestion
        BigInteger latestBlockNumberWithDealBigInt = dealWatcherService.getLatestBlockNumberWithDeal();
        latestBlockNumberWithDeal = Metrics.gauge("iexec.core.deals.last.block", new AtomicLong(latestBlockNumberWithDealBigInt == null ? -1L : latestBlockNumberWithDealBigInt));
```

The var name may not be the best choice there ;)

src/test/java/com/iexec/core/metric/MetricServiceTests.java
```diff
@@ -78,4 +91,46 @@ void shouldGetPlatformMetrics() {
...
+    @Test
+    void shouldInitializeMetrics() {
+        verifyGaugeMetric("iexec.core.workers", 3);
+        verifyGaugeMetric("iexec.core.cpu.total", 1);
+        verifyGaugeMetric("iexec.core.cpu.available", 1);
+        verifyGaugeMetric("iexec.core.tasks.completed", 0);
+        verifyGaugeMetric("iexec.core.deals.events", 10L);
+        verifyGaugeMetric("iexec.core.deals", 8);
+        verifyGaugeMetric("iexec.core.deals.replay", 2L);
+        verifyGaugeMetric("iexec.core.deals.last.block", 255L);
```
mcornaton, line 103: These magic numbers could be converted to constants. It may make those tests more easy to maintain and update later on.

src/main/resources/application.yml
thewhitewizard: I don't think we should make this a configurable choice. Any administrator wants to be able to monitor how the application is working.

src/test/java/com/iexec/core/metric/MetricServiceTests.java
thewhitewizard: public static final String METRIC_WORKERS = "iexec.core.workers";
    public static final String METRIC_CPU_TOTAL = "iexec.core.cpu.total";
    public static final String METRIC_CPU_AVAILABLE = "iexec.core.cpu.available";
    public static final String METRIC_TASKS_COMPLETED = "iexec.core.tasks.completed";
    public static final String METRIC_DEALS_EVENTS = "iexec.core.deals.events";
    public static final String METRIC_DEALS_COUNT = "iexec.core.deals";
    public static final String METRIC_DEALS_REPLAY = "iexec.core.deals.replay";
    public static final String METRIC_DEALS_LAST_BLOCK = "iexec.core.deals.last.block";

src/main/java/com/iexec/core/metric/MetricService.java
thewhitewizard: I can cheat and rename all my gauges with the metric suffix, so I can use the correct variable name

src/main/resources/application.yml
jbern0rd: confusion here, this will expose `/actuator/metrics` and has nothing to do with `/metrics`

src/main/java/com/iexec/core/metric/MetricService.java
thewhitewizard: or maybe like that

        latestBlockNumberWithDeal = Metrics.gauge(METRIC_DEALS_LAST_BLOCK,
                new AtomicLong(Optional.ofNullable(dealWatcherService.getLatestBlockNumberWithDeal()).orElse(BigInteger.valueOf(-1)).longValue()));

jbern0rd: Something is looking weird to me here. `latestBlockNumberWithDeal` is initialized at `BigInteger.ZERO`.
I have the impression it should never be `null`. By convention, our APIs seems to return 0 instead of -1 when the blockchain is not reachable.
For those kind of metrics, why don't we just expose them in the correct service and scrap them to create exposed metric when the MetricsController is actually called.

For example, I think the latest block number metric should be managed in DealWatcherService. It should be updated every time a valid deal is received and be fetched by metrics collection service when they perfom their scrape action.
What do you think ?

src/main/resources/application.yml
mcornaton: That's it. `/actuator/metrics` will in fact expose default metrics (disk free/used/total, heap used, ...) and custom metrics ("iexec.core.workers", ...).
So, that's quite dangerous to expose it as a default. But `/metrics` is already exposed, so admin can access those values anyway :)

src/test/java/com/iexec/core/metric/MetricServiceTests.java
mcornaton: I was also talking about the values (3, 1, 1, 0, 10L, 8, 2L, 255L) :)

src/main/java/com/iexec/core/worker/WorkerService.java
```diff
@@ -50,15 +60,47 @@ public WorkerService(WorkerRepository workerRepository,
         this.contextualLockRunner = new ContextualLockRunner<>();
     }
 
+    @PostConstruct
+    void init() {
+        aliveWorkersGauge = Metrics.gauge(METRIC_WORKERS_GAUGE, new AtomicInteger(getAliveWorkers().size()));
+        aliveTotalCpuGauge = Metrics.gauge(METRIC_CPU_TOTAL_GAUGE, new AtomicInteger(getAliveTotalCpu()));
+        aliveAvailableCpuGauge = Metrics.gauge(METRIC_CPU_AVAILABLE_GAUGE, new AtomicInteger(getAliveAvailableCpu()));
```
jbern0rd, line 67: I think I would prefer to have an init at `0` here. Then, `true` values will be populated by the scheduled mechanism.
It would feel weird to me to see a scheduler starting with `10` workers to discover that only `4` are actually connected.

src/main/java/com/iexec/core/worker/WorkerService.java
```diff
@@ -50,15 +60,47 @@ public WorkerService(WorkerRepository workerRepository,
...
+    }
+
+    /**
+     * updateMetrics is used to update all workers metrics
+     */
+    @Scheduled(fixedDelayString = "${cron.metrics.refresh.period}", initialDelayString = "${cron.metrics.refresh.period}")
```
jbern0rd, line 73: Can we have a following PR with an updated `README.md` containing the new parameters ?
It would be nice to add a section like the one recently added on the worker to let users know they can enable prometheus metrics.

jbern0rd: I think that in a next PR, I would like to read the Prometheus metrics value when calling `/metrics` endpoint.
This would avoid some non-optimized calls on the database for example.

src/main/java/com/iexec/core/worker/WorkerService.java
jbern0rd: After some more thinking, the proposed implementation is really accurate. It will fetch the correct number of workers after a scheduler restart. If the scheduler was offline for a long time, no alive workers will be seen.

src/main/java/com/iexec/core/metric/MetricService.java
```diff
@@ -20,11 +20,12 @@
 import com.iexec.core.task.TaskService;
 import com.iexec.core.task.TaskStatus;
 import com.iexec.core.worker.WorkerService;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
 @Service
+@Slf4j
```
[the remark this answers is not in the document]
thewhitewizard: remove in local

src/main/java/com/iexec/core/worker/WorkerService.java
thewhitewizard: so nothing to do on my side ?

jbern0rd: nope, forget about it

src/main/resources/application.yml
jbern0rd: do you still wish to enable `/actuator/metrics` by default ? This configuration has nothing to do with our `/metrics` endpoint which is always enabled

PR #637, opened 2023-12-01
Expose version through prometheus endpoint

thewhitewizard:


src/main/java/com/iexec/core/config/ObservabilityConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public static final String METRIC_INFO_GAUGE_NAME = "iexec.app.info";
+    public static final String METRIC_INFO_GAUGE_DESC = "A metric to expose version and application name.";
+    public static final String METRIC_INFO_LABEL_APP_NAME = "app_name";
+    public static final String METRIC_INFO_LABEL_APP_VERSION = "app_version";
+
+    ObservabilityConfiguration(@Autowired BuildProperties buildProperties) {
+
+        Gauge.builder(METRIC_INFO_GAUGE_NAME, 1.0, n -> n)
+                .strongReference(true)
```
mcornaton, line 36: Is this strong reference required?

src/test/java/com/iexec/core/config/ObservabilityConfigurationTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @AfterEach
+    void afterEach() {
+        Metrics.globalRegistry.clear();
+    }
+
+    @BeforeEach
+    void init() {
+        MockitoAnnotations.openMocks(this);
+    }
```
mcornaton, line 50: Can we reorder these methods? I would use the following order:
```java
@BeforeAll
@BeforeEach
@AfterEach
@AfterAll
```

src/test/java/com/iexec/core/config/ObservabilityConfigurationTest.java
```diff
@@ -0,0 +1,68 @@
...
+        Mockito.when(buildProperties.getVersion()).thenReturn(version);
+        Mockito.when(buildProperties.getName()).thenReturn(name);
+
+        final ObservabilityConfiguration observabilityConfiguration = new ObservabilityConfiguration(buildProperties);
+        Assertions.assertThat(observabilityConfiguration).isNotNull();
+        final Gauge info = Metrics.globalRegistry.find(ObservabilityConfiguration.METRIC_INFO_GAUGE_NAME).gauge();
+        Assertions.assertThat(info).isNotNull();
+        Assertions.assertThat(info.getId()).isNotNull();
+        Assertions.assertThat(info.getId().getTag(ObservabilityConfiguration.METRIC_INFO_LABEL_APP_NAME)).isEqualTo(name);
+        Assertions.assertThat(info.getId().getTag(ObservabilityConfiguration.METRIC_INFO_LABEL_APP_VERSION)).isEqualTo(version);
```
mcornaton, line 65: Can we use an `assertAll`?

thewhitewizard: as one of the great presidents said, yes we can

jbern0rd: Do we really need to mock things here ?
Why not simply assert we have `BuildProperties` with correct values ?

src/main/java/com/iexec/core/config/ObservabilityConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+package com.iexec.core.config;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.Metrics;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.info.BuildProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ObservabilityConfiguration {
```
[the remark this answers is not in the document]
thewhitewizard: If you want, I can:

Add a getAppName in VersionService and add a postConstruct in the Controller to create the metric (by calling VersionService).

OR

Keep only the controller (delete the service which is a bit too much here)


But I liked the idea of managing this metric independently in a Config class because it can be replicated in all projects (we don't necessarily have the controller everywhere).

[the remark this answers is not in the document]
jbern0rd: The idea for me is to have the controller everywhere.
It will be more work but every app should have the version controller and the version metric endpoint in my opinion.

jbern0rd (approved):
Do we want to try to update `springdoc-openapi` dependency to the highest possible version  (1.7.0 as we still use Spring Boot 2.x https://springdoc.org/) ?

PR #641, opened 2023-12-11
Use `BlockchainAdapterService` from `iexec-blockchain-adapter-api-library`

jbern0rd:


src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -73,14 +76,14 @@ public TaskUpdateManager(TaskService taskService,
...
-                             BlockchainAdapterService blockchainAdapterService,
+                             BlockchainAdapterApiClient blockchainAdapterApiClient,
                              SmsService smsService) {
         this.taskService = taskService;
         this.iexecHubService = iexecHubService;
         this.replicatesService = replicatesService;
         this.applicationEventPublisher = applicationEventPublisher;
         this.workerService = workerService;
-        this.blockchainAdapterService = blockchainAdapterService;
+        this.blockchainAdapterService = new BlockchainAdapterService(blockchainAdapterApiClient, WATCH_PERIOD_SECONDS, MAX_ATTEMPTS);
```
mcornaton, line 86: Why don't you inject the bean instead of creating it here?

jbern0rd: It is the only place where it is used and it allowed me to pass the 80% coverage threshold.
I don't really know the best solution between single use spring beans or traditional java object instances creation for this kind of situation.

jbern0rd: reverted in last commits

mcornaton (approved):

thewhitewizard (approved):

PR #663, opened 2024-02-12
Fix enclave challenge generation failure.

mcornaton:


src/main/java/com/iexec/core/sms/SmsService.java
```diff
@@ -119,27 +117,25 @@ public Optional<String> getEnclaveChallenge(String chainTaskId, String smsUrl) {
...
+                log.error("An error occurred while getting teeChallengePublicKey "
+                        + "[chainTaskId:{}, smsUrl:{}]", chainTaskId, smsUrl);
+                return Optional.empty();
+            }
 
-        if (StringUtils.isEmpty(teeChallengePublicKey)) {
-            log.error("An error occurred while getting teeChallengePublicKey " 
-                + "[chainTaskId:{}, smsUrl:{}]", chainTaskId, smsUrl);
+            return Optional.of(teeChallengePublicKey);
+        } catch (FeignException e) {
```
jbern0rd, line 136: With a reverse proxy, other runtime exceptions could happen in production.
Do we want to have the biggest net and catch all exceptions here ?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 
 - Filter out `CONTRIBUTE_AND_FINALIZE` tasks when detecting missed `REVEALED` status update. (#658)
 - Fetch `results` on-chain when updating a replicate status in `CONTRIBUTE_AND_FINALIZE` workflow. (#659 #660)
+- Fix enclave challenge generation failure. (#663)
```
jbern0rd, line 15: Properly catch all exceptions when an enclave challenge generation fails ?
The idea behind this suggestion is to give an idea of the fix

src/main/java/com/iexec/core/sms/SmsService.java
mcornaton: Thanks for the idea. Here is a commit addressing the issue: https://github.com/iExecBlockchainComputing/iexec-core/pull/663/commits/a8997d4d9aca42f3c13989da2da320a2f9f2b8a6
Catching unchecked exceptions should be enough there, as Feign methods contract do not allow them to throw checked exceptions.

CHANGELOG.md
mcornaton: Sure, please see https://github.com/iExecBlockchainComputing/iexec-core/pull/663/commits/a8997d4d9aca42f3c13989da2da320a2f9f2b8a6.

jbern0rd (approved):

PR #675, opened 2024-03-15
Add logs for better traceability

thewhitewizard:


src/main/java/com/iexec/core/worker/WorkerController.java
```diff
@@ -125,6 +130,7 @@ public ResponseEntity<Worker> registerWorker(@RequestHeader("Authorization") Str
 
     @GetMapping(path = "/workers/config")
     public ResponseEntity<PublicConfiguration> getPublicConfiguration() {
+        log.debug("A worker asks for the public configuration");
```
jbern0rd, line 133: Maybe something more generic, it may not be a worker

src/main/java/com/iexec/core/worker/WorkerUtils.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.iexec.core.worker;
+
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+
+@Slf4j
+public class WorkerUtils {
```
jbern0rd, line 23: Can we add `@NoArgsConstructor(access = AccessLevel.PRIVATE)` to have an happy sonarcloud ?

thewhitewizard: as a famous president said, yes we can

jbern0rd (approved):

PR #695, opened 2024-04-18
Create `iexec-task-api` to access task feedback API

jbern0rd:


iexec-core-library/src/main/java/com/iexec/core/api/SchedulerClient.java
thewhitewizard, line 1: the year update in the header is missing

thewhitewizard (approved):

PR #696, opened 2024-04-22
Rename `IexecTaskApiClient` to `TaskApiClient`

jbern0rd:


iexec-core-library/src/main/java/com/iexec/core/api/SchedulerClient.java
```diff
@@ -58,7 +58,7 @@ public interface SchedulerClient {
 
     @RequestLine("POST /workers/register")
     @Headers("Authorization: {authorization}")
-    void registerWorker(@Param("authorization") String authorization, WorkerModel model);
+    Void registerWorker(@Param("authorization") String authorization, WorkerModel model);
```
mcornaton, line 61: Is there a reason why we need a `Void` return type there?

jbern0rd: This change is due to generic use in worker

thewhitewizard (approved):

PR #701, opened 2024-05-22
Deprecate legacy task feedback API endpoints

jbern0rd:


src/test/java/com/iexec/core/task/TaskControllerTests.java
```diff
@@ -86,11 +87,29 @@ void setUp() {
...
-        return Credentials.create(ecKeyPair).getAddress();
+        final ECKeyPair otherKeyPair = Keys.createEcKeyPair();
+        return Credentials.create(otherKeyPair).getAddress();
     }
     //endregion
 
+    // region deprecated methods, to remove in future version
+    @Test
+    void shouldCallGetTaskLogs() {
+        final TaskController controller = spy(new TaskController(challengeService, iexecHubService, replicatesService, taskLogsService, taskService));
```
mcornaton, line 98: Is there a reason why you create a new `TaskController` instead of spying on the already-existing one?
```suggestion
        final TaskController controller = spy(taskController);
```

thewhitewizard (approved):

src/test/java/com/iexec/core/task/TaskControllerTests.java
jbern0rd: fixed, thanks for the suggestion

mcornaton (approved):

PR #702, opened 2024-05-23
Do not supply replicates past their contribution deadlines to workers

jbern0rd:


thewhitewizard (approved):

src/main/java/com/iexec/core/task/TaskRepository.java
```diff
@@ -39,22 +39,23 @@ public interface TaskRepository extends MongoRepository<Task, String> {
...
+     * @param timestamp            The task contribution deadline should be after the provided timestamp.
      * @param excludedTags         The task tag should not be one this tag list
      *                             - use {@literal null} if no tag should be excluded.
      * @param excludedChainTaskIds The chain task ID should not be one of this list.
      * @param sort                 How to prioritize tasks.
      * @return The first task matching with the criteria, according to the {@code sort} parameter.
      */
-    Optional<Task> findFirstByCurrentStatusInAndTagNotInAndChainTaskIdNotIn(List<TaskStatus> statuses, List<String> excludedTags, List<String> excludedChainTaskIds, Sort sort);
+    Optional<Task> findFirstByCurrentStatusInAndContributionDeadlineAfterAndTagNotInAndChainTaskIdNotIn(
+            List<TaskStatus> statuses, Date timestamp, List<String> excludedTags, List<String> excludedChainTaskIds, Sort sort);
```
mcornaton, line 58: Are you sure using an old `Date` instance is safe? I would recommend using a `LocalDateTime` if this works.

jbern0rd: The database schema still uses date. I am not confident in modifying it in the PR. It may require a dedicated activity to modify this where need while ensuring there is no breaking change.

mcornaton (approved):

PR #718, opened 2024-11-14
Push scheduler Result Proxy URL as Web2 secret to SMS

jbern0rd:


src/main/java/com/iexec/core/sms/SmsService.java
```diff
@@ -16,35 +16,59 @@
...
+                      final SignatureService signatureService,
+                      final SmsClientProvider smsClientProvider) {
+        this.iexecHubService = iexecHubService;
         this.registryConfiguration = registryConfiguration;
+        this.resultRepositoryConfiguration = resultRepositoryConfiguration;
         this.signatureService = signatureService;
         this.smsClientProvider = smsClientProvider;
     }
 
+    private Optional<String> getVerifiedSmsUrl(String chainTaskId) {
```
nabil-Tounarti, line 63: Wouldn't it be better to add final to the chainTaskId parameter for consistency? The same applies to chainTaskId and tag parameters in the getVerifiedSmsUrl function.

jbern0rd: Thanks, fixed in 174bed6

src/main/java/com/iexec/core/sms/SmsService.java
```diff
@@ -147,4 +170,51 @@ Optional<String> generateEnclaveChallenge(String chainTaskId, String smsUrl) {
...
+    }
+
+    /**
+     * Pushes worker pool default Result Proxy URL to SMS.
+     *
+     * @param event The default result-proxy URL.
+     */
+    @EventListener
+    public void pushWorkerpoolResultProxyUrl(final TaskInitializedEvent event) {
+        log.debug("pushWorkerpoolResultProxyUrl");
```
Natchica, line 199: Is this there on purpose (if yes, shouldn't it be a bit more descriptive ?) or did you forget to delete it ?
```suggestion
        log.debug("Call to pushWorkerpoolResultProxyUrl(event) [event: {}]", event);
```

jbern0rd: This is an habit of mine, having a debug log message at the beginning of methods to see how methods are chained.
I don't really like the **Call to** prefix which will be added in many places for little added-value.
I will try the end of your suggestion to see the produced output.

jbern0rd: see 4af05bf

nabil-Tounarti (approved):

thewhitewizard (approved):

PR #720, opened 2024-11-27
Update off-chain task model with on-chain task consensus data in all workflows

jbern0rd:


src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -106,7 +109,7 @@ void updateTask(String chainTaskId) {
                 initialized2Running(task, chainTask);
                 break;
             case RUNNING:
-                transitionFromRunningState(task, chainTask);
+                transitionFromRunningState(chainTask, task);
```
Natchica, line 112: Why did you switch the two parameters?

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -348,7 +353,18 @@ void initialized2Running(Task task, ChainTask chainTask) {
...
+    /**
+     * Handle transition when an off-chain task in the {@code RUNNING} state.
+     * <p>
+     * This method handles contribution timeouts, whether replicates list is empty and task transition to the next state
+     * depending on the task eligibility to {@code contributeAndFinalize} flow.
+     *
+     * @param chainTask On-chain task
+     * @param task      Off-chain task model of the on-chain task
+     */
+    void transitionFromRunningState(final ChainTask chainTask, final Task task) {
```
Natchica, line 367: Same question as my above comment (Why not just add the `final` and keep the original params order ?)

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
```diff
@@ -377,9 +393,9 @@ void transitionFromRunningState(Task task, ChainTask chainTask) {
             // running2Finalized2Completed must be the first call to prevent other transition execution
             log.debug("Task is running in a TEE without callback, flow contributeAndFinalize is possible [chainTaskId:{}]",
```
Natchica, line 394: Shouldn't this log debug message be updated like the comment on line 398 (ex 382) ?

Natchica (commented):
Nice work on the added documentation !

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
jbern0rd: All methods called from this one have the `(chainTask, task)` order.
As data are copied from `chainTask` to `task`, I told myself it could be more maintainable to have the same order while keeping the notion of `source -> target`.

src/main/java/com/iexec/core/task/update/TaskUpdateManager.java
jbern0rd: fixed in 7e16bda

nabil-Tounarti (approved):

Natchica (approved):

